PR #24, opened 2023-03-14
Add LIKE and ILIKE operators

isgj:


stephenafamo:
ILIKE is not part of the SQL spec and is not implemented in SQLite.

So I would suggest keeping `LIKE` in the `expr` package and adding `ILIKE` as a specific Postgres and MySQL operator.

isgj:
I removed `ILIKE` from the `expr`.

Not sure where to put ILIKE (straters.go???), also it will be just a wrapper of `expr.OP`.

What do you think of adding `WhereMod.LIKE` (`WhereMod.ILIKE`)?

stephenafamo:
> Not sure where to put ILIKE (straters.go???)

In `dialect/builder.go`. Postgres currently has `BetweenSymmetric`

> What do you think of adding `WhereMod.LIKE` (`WhereMod.ILIKE`)?

I like it! Makes sense.

Some more things. 
1. I think these operators should take a string instead of `any` because the operator is meant to take a pattern string.
2. Since `LIKE` is usually used with `%`. It may be convenient to add methods to add `%` as a prefix, suffix or both.

stephenafamo:
> * I think these operators should take a string instead of `any` because the operator is meant to take a pattern string.

Actually..... while `LIKE` expects a string pattern, that can be supplied by any Postgres expression. So instead, let it accept a `bob.Expression` similar to other operators like `EQ`

isgj:
> Since LIKE is usually used with %. It may be convenient to add methods to add % as a prefix, suffix or both.

Added the utility methods `StartsWith`, `Contains`, `EndsWith`

stephenafamo:
> Added the utility methods `StartsWith`, `Contains`, `EndsWith`

Thank you so much. But actually thinking about this more, I no longer think this is the right thing to do.

While convenient, it kind of goes against the project's principle of following the spec as closely as possible.  
For example, PostgreSQL actually has an inbuilt `starts_with` function and having a method with that name could make things confusing.

Finally, it is much easier to add new methods in the future rather than remove existing ones, so maybe it is better to not yet add convenience functions.

I apologise for asking you to add this and now asking you to remove it 😮‍💨 .

isgj:
sure no problem. It makes sense

stephenafamo (approved):

PR #26, opened 2023-03-18
docs: typo, remove brackets

leonardtan13:


stephenafamo:
This is incorrect. As a matter of fact, a context is supposed to be passed in the brackets.

Issue #34, opened 2023-03-28
Embed structs


raphaelvigee:
If i define the following:

```
type Flow struct {
	*models.Flow
	Model `db:"-"`
}
```

where `models` is the package generated models, and Model is a base utility that adds some Methods to my Flow structure, i m unable to Scan into it because the `internal.GetMappings` does not read embeded structs fields and only see a single column: `flow` (whic is not correct), it would be great to have it inline embeded struct fields

This method allows to use the generated models and augment them with custom business logic

stephenafamo:
It does not read the mappings because you set the DB tag as `db:"-"`

stephenafamo:
So, to do this, we may need to have a special DB tag to make the `GetMappings` method know what to look for.

raphaelvigee:
Sorry I wasnt super clear, `Model` is a purely go interface stuff, must be ignored by the DB scan, hence the `-`, `*models.Flow` on the other hand contains the actual fields that need to be scanned into,

Running `internal.GetMappings`  on the `Flow` type gives out a single column (and mapper): `flow` (which maps to the embeded struct), where i expect it to list the columns contained inside the embeded struct `models.Flow`

stephenafamo:
I'll try to think of a way to achieve this.


In the meantime, perhaps you can consider extending the templates to add shared logic that may be more elegant than embedding

raphaelvigee:
Tbh i ve been thinking that extending the template would be a lot better, but couldnt find any details on how to do that, do you have any docs/example on how to achieve that ?

stephenafamo:
No docs yet, but looking at the drivers themselves should help. A lot is abstracted. For example the `main` function of `bobgen-psql` is only [15 lines](https://github.com/stephenafamo/bob/blob/main/gen/bobgen-psql/main.go#LL45-L60C2).

raphaelvigee:
So its down to making a custom gen binary similar to what ent is doing? Awesome then :) Thx for this great lib bth!

stephenafamo:
I'm currently using a similar system to add GraphQL Relay semantics to a project of mine (Universal NodeIDs, Connections i.e. cursor pagination), hoping to release that as a standalone package soon.

A few things may change about it to be fair. I'm considering having more granular plugin hooks similar to `gqlgen`, but haven't settled on a proper implementation yet.

stephenafamo:
> Thx for this great lib btw!

Thanks a lot 😁

[...]

PR #37, opened 2023-03-30
Add additional methods hook

raphaelvigee:
ref #34

### DIFF
--- a/gen/templates/models/01_types.go.tpl
+++ b/gen/templates/models/01_types.go.tpl
@@ -28,6 +28,8 @@ type {{$tAlias.UpSingular}} struct {
 	{{end -}}
 }
 
+{{block "model/methods/additional" .}}{{end}}
+
 // {{$tAlias.UpSingular}}Slice is an alias for a slice of pointers to {{$tAlias.UpSingular}}.
 // This should almost always be used instead of []{{$tAlias.UpSingular}}.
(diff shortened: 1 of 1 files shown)


stephenafamo:
Is this hook necessary? One can just define the methods in the additional templates and they will be rendered at the end of the file.

raphaelvigee:
Oh really ? But how do you tell which file it will end up in ?

raphaelvigee:
Also can confirm, extra stuff in the template ends up being in another file in the generated directory (in my case `gen/template/<model>.go`)
So i would say yes, required?!

stephenafamo:
The directory of the templates should be like this

```
templates/
|-- singleton/
|-- singleton/somthing.go.tpl 
|-- another.go.tpl
```

Files in singleton (e.g. `something.go.tpl`) are run once and generated as-is.  
Files in the root directory(e.g. `another.go.tpl`) are run for each table/view and appended to `model.go`

If you look in `gen/templates` you can see the structure.

Note, if you're using `embed`, it will add the name of the directory to the templates which sometimes messes with how they are interpreted which is why if you look in [`templates.go`](https://github.com/stephenafamo/bob/blob/main/gen/templates.go#LL33-L39C2) you see that I use `fs.Sub` to change the root of the `embed.Fs`

raphaelvigee:
Right got it, not required then, thx!

PR #66, opened 2023-06-07
Named arguments

RangelReale:
Creates named arguments at generation time and allows replacements using a map/struct.


sample/sample.go
```diff
@@ -0,0 +1,196 @@
+package main
```
RangelReale, line 1: this file is temporary for showing the usage, will be removed in the final version.

query_build.go
```diff
@@ -0,0 +1,84 @@
...
+		} else {
+			hasNonNamed = true
+		}
+	}
+	if len(nargs) == 0 {
+		return args, nil
+	}
+	if hasNonNamed {
+		return nil, fmt.Errorf("cannot mix named and non-named arguments")
+	}
```
RangelReale, line 23: tecnically it wouldn't be a problem mixing named and non-named parameters, I could just forward the parameter in the same order if it wasn't a `NamedArgument`.

stephenafamo: But how will that work with prepared statements?

RangelReale: `mergeNamedArguments` would receive `[]any`, and in [this for loop](https://github.com/stephenafamo/bob/pull/66/files#diff-a1785b1d4282adef1513787b42f45fb946e31f6f6a3c9c4953a50c8e4302df6cR38), I would check if the type is `NamedArgument`, otherwise I would just add the value itself to the return array at the same position.

RangelReale: for example, given:

```go
	query := psql.Insert(
		im.Into("actor", "first_name", "last_name"),
                im.Values(psql.Arg(12, psql.NamedArg("in"))),
	)
```

`mergeNamedArguments` would receive as first parameter:

`[]any{12, NamedArg{"in"}}`

in the for loop I would just forward the value that is not NamedArg and only replace the other one, so in:

```go

	queryStr, args, err := bob.BuildWithNamedArgs(query, map[string]any{
		"in": "LAST_NAME",
	})
```

`args` would be `[]any{12, "LAST_NAME"}`.

dialect/psql/starters.go
```diff
@@ -62,18 +62,30 @@ func Arg(args ...any) Expression {
 	return bmod.Arg(args...)
 }
 
+func ArgNamed(names ...string) Expression {
+	return bmod.ArgNamed(names...)
+}
```
RangelReale, line 67: In the discussion you talked about calling this `NamedArg`, but there is a `NamedArg` function already in this scope, maybe I could call that one `Named`, which would more or less resemble the `sql.Named` function which is for named arguments also?

RangelReale:
One thing that I'm not entirelly sure is the "Named args" naming, I think it may confuse users about SQL servers that do support named parameters.

Maybe we could call it "Bind" or "Bound" instead of "Named"? `BoundArgument`, `BindArg`, etc.

RangelReale:
I did a quick test using "Bind" or "Binding" instead of "named arg", and I think it made the intent much clearer.
[RangelReale:bind-args-v3](https://github.com/stephenafamo/bob/compare/main...RangelReale:bind-args-v3)

With this naming it was also possible to make it much simpler, instead of new top-level functions, a single `BindQuery` toplevel function which returns an interface that extends `Query` and also has the 4 `Build` functions made it work similar to what is done currently.

Also adding a `.Bind` method to `BaseQuery` made it much easier to use, for example:

```go
	queryStr, args, err := query.Bind(map[string]any{
		"in1": 15,
		"in2": "LAST_NAME",
	}).Build()
```

stephenafamo:
See #104

PR #68, opened 2023-06-22
Bind arguments (alternative to Named arguments)

RangelReale:
Creates bound arguments at generation time and allows replacements using a map/struct.

Alternative syntax for #66


stephenafamo:
See #104

PR #78, opened 2023-07-17
Always surround subqueries with parenthesis when used as an expression

### LINKED ISSUE #76: Subqueries in select lists should be wrapped in parenthesis
This is a [documentation I found about it](https://www.w3resource.com/PostgreSQL/postgresql-subqueries.php).

```
A subquery must be enclosed in parentheses. 
```

```go
psql.Select(
    sm.Columns("id"),
    sm.Columns(psql.Select(
        sm.Columns("address"),
        sm.From("users_address ua"),
        sm.Where("users.id = ua.id"),
    )),
    sm.From("users"),
    sm.Where(psql.Quote("id").In(psql.Arg(100, 200, 300))),
)
```

output query:

```sql
SELECT 
id, SELECT 
address
FROM users_address ua
WHERE users.id = ua.id

FROM users
WHERE ("id" IN ($1, $2, $3))
```

stephenafamo:
Fixes #76

### DIFF
--- a/clause/from.go
+++ b/clause/from.go
@@ -99,16 +99,9 @@ func (f From) WriteSQL(w io.Writer, d bob.Dialect, start int) ([]any, error) {
 		w.Write([]byte("LATERAL "))
 	}
 
-	_, isQuery := f.Table.(bob.Query)
-	if isQuery {
-		w.Write([]byte("("))
-	}
 	args, err := bob.Express(w, d, start, f.Table)
 	if err != nil {
 		return nil, err
-	}
-	if isQuery {
-		w.Write([]byte(")"))
 	}
 
 	if f.WithOrdinality {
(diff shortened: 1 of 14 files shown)


RangelReale:
This change made `psql.Where(psql.RawQuery("xxx"))` to always surround the condition in parenthesis even if not needed, which it wasn't doing before. Not a deal breaker, but queries are easier to read with less parenthesis.

RangelReale:
This also happens on `JOIN table ON (condition)`, it now always have parenthesis when using `RawQuery` or `Raw`.

stephenafamo:
> This change made `psql.Where(psql.RawQuery("xxx"))` to always surround the condition in parenthesis

I believe `psql.Raw` should be used here instead. `RawQuery` is expected to be used on full queries, e.g. `SELECT * FROM table_name`

PR #79, opened 2023-07-17
fix: do not add alias in update/delete all queries

### LINKED ISSUE #75: Questions about generated query builders
The generated query builder function only take `bob.Mod[dialect.SelectQuery]`, but returns `TableQuery`, and methods like `UpdateAll` are still available on the type. For example:

```go
	models.Videos(ctx, db).UpdateAll(&models.VideoSetter{
		SponsorID: omitnull.From(1),
	})
```

will result in the following query and error:

```
UPDATE `videos` AS `videos` SET
`sponsor_id` = ?
WHERE ((`id`) IN (SELECT
`id`
FROM `videos` AS `videos`
))
0: omitnull.Val[int]{value:1, state:2}

Error 1093 (HY000): You can't specify target table 'videos' for update in FROM clause
```

While this is understandable, it doesn't quite match the 'find error in compile time' design style as in separating sm/im/um packages for query builder. Maybe we can make the methods on query more restricted by having more types of query? e.g. `TableSelectQuery`, `TableUpdateQuery`.

Also I couldn't find any genereated code (without using raw dialect specific query builder) as alternative to the following sqlboiler experience:

```go
models.Videos(mods...).UpdateAll(ctx, db, models.M{models.VideoColums.SponsorID: 1})
```

I end up generating code like this (using the test database in bob as example):

usage:

```go
	rowsAffected, err := models.UpdateVideos(
		ctx, bob.Debug(db),
		models.UpdateWhere.Videos.UserID.GT(10),
	).Set(models.VideoSetter{
		SponsorID: omitnull.From(1),
	})
```

in singleton:

```go
type updateQueryMod interface {
	applyUpdateQuery(q *dialect.UpdateQuery) int
}

type UpdateQuery[Tset updateQueryMod] struct {
	q bob.BaseQuery[*dialect.UpdateQuery]
	ctx context.Context
	exec bob.Executor
}

func (uq *UpdateQuery[Tset]) Set(s Tset) (int64, error) {
	if s.applyUpdateQuery(uq.q.Expression) == 0 {
		return 0, nil
	}
	query, args, err := uq.q.Build()
	if err != nil {
		return 0, err
	}
	res, err := uq.exec.ExecContext(uq.ctx, query, args...)
	if err != nil {
		return 0, err
	}
	return res.RowsAffected()
}
```

in model:

```go
func (vs VideoSetter) applyUpdateQuery(q *dialect.UpdateQuery) (applied int) {
	if vs.ID.IsSet() {
		um.Set(TableNames.Videos, ColumnNames.Videos.ID).ToArg(vs.ID).Apply(q)
		applied++
	}
	if vs.UserID.IsSet() {
		um.Set(TableNames.Videos, ColumnNames.Videos.UserID).ToArg(vs.UserID).Apply(q)
		applied++
	}
	if vs.SponsorID.IsSet() {
		um.Set(TableNames.Videos, ColumnNames.Videos.SponsorID).ToArg(vs.SponsorID).Apply(q)
		applied++
	}
	return applied
}

func UpdateVideos(
	ctx context.Context, 
	exec bob.Executor, 
	mods ...bob.Mod[*dialect.UpdateQuery],
) *UpdateQuery[VideoSetter] {
	q := mysql.Update(um.Table(TableNames.Videos))
	q.Apply(mods...)
	return &UpdateQuery[VideoSetter]{
		q: q,
		ctx: ctx,
		exec: exec,
	}
}
```

I wonder if there's a better way to support these types for query out of the box with bob.

stephenafamo:
Fixes #75

### DIFF
--- a/dialect/mysql/table.go
+++ b/dialect/mysql/table.go
@@ -524,7 +524,7 @@ func (t *Table[T, Tslice, Tset]) Query(ctx context.Context, exec bob.Executor, q
 	vq := t.View.Query(ctx, exec, queryMods...)
 	return &TableQuery[T, Tslice, Tset]{
 		ViewQuery:  *vq,
-		nameExpr:   t.NameAs,
+		nameExpr:   t.Name,
 		pkCols:     t.pkCols,
 		setMapping: t.setMapping,
 	}
@@ -532,7 +532,7 @@ func (t *Table[T, Tslice, Tset]) Query(ctx context.Context, exec bob.Executor, q
 
 type TableQuery[T any, Ts ~[]T, Tset any] struct {
 	ViewQuery[T, Ts]
-	nameExpr   func(context.Context) bob.Expression // to prevent calling it prematurely
+	nameExpr   func(context.Context) dialect.Expression // to prevent calling it prematurely
 	pkCols     []string
 	setMapping internal.Mapping
 }
--- a/dialect/psql/table.go
+++ b/dialect/psql/table.go
@@ -470,7 +470,7 @@ func (t *Table[T, Tslice, Tset]) Query(ctx context.Context, exec bob.Executor, q
 	vq := t.View.Query(ctx, exec, queryMods...)
 	return &TableQuery[T, Tslice, Tset]{
 		ViewQuery:  *vq,
-		nameExpr:   t.NameAs,
+		nameExpr:   t.Name,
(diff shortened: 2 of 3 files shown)


stephenafamo:
This does not actually fix #75 since it is caused by a quirk of MySQL.

PR #80, opened 2023-07-18
Make model hooks take slices not single objects

### LINKED ISSUE #74: Different hooks for One and Many operations
I noticed that bob uses the same hook for both One and Many operations. Like here in `AfterUpdateHook`:

https://github.com/stephenafamo/bob/blob/4824718378d4e83558ecb8f5ff7a1f94a4bf1d99/dialect/mysql/table.go#L296-L301

I think it might be desirable in quite some cases to handle all rows Updated/Deleted/Inserted together, e.g. update another table in batch, record the number of rows into some kind of metrics. And passing an executor to a function in a loop might lead to unexpected performance pitfalls.

Maybe we can have something like `AfterUpdateManyHook`, `AfterUpdateOneHook` ? It's likely easy to implement but keeping backward compatibility is quite tricky.

A not very elegant way I can think of is abusing the context parameter. Add a pairs of functions
- `orm.RequestBatchHook(context.Context) context.Context` 
- `orm.BatchHookRequested(context.Context) bool`. 

The user who wanted to handle hooks in batch adds an `AfterUpdateHook` in which `orm.RequestBatchHook` is called. Then add their hook function that handles multiple rows in an `AfterUpdateManyHook`.

In corresponding methods in `Table`, `AfterUpdateManyHook` is used if requested.

```go
	for i, row := range rows {
		ctx, err = t.AfterUpdateHooks.Do(ctx, exec, row)
		if err != nil {
			return 0, err
		}
                 if orm.BatchHookRequested(ctx) && i == 0 {
                         if _, err := t.AfterUpdateManyHooks.Do(ctx, exec, rows); err != nil {
                                 return 0, err
                         }
                         break
                 }
	}
```

Though I'm not quite sure if it's worth adding this complexity. What can be done by hooks can always be done manually without hooks. Any suggestions ?

stephenafamo:
This should fix #74 

Take a look and let me know what you think @inoc603

### DIFF
--- a/dialect/mysql/table.go
+++ b/dialect/mysql/table.go
@@ -57,19 +57,19 @@ type Table[T any, Tslice ~[]T, Tset any] struct {
 	pkCols     []string
 	setMapping internal.Mapping
 
-	BeforeInsertHooks orm.Hooks[Tset]
-	BeforeUpsertHooks orm.Hooks[Tset]
+	BeforeInsertHooks orm.Hooks[[]Tset]
+	BeforeUpsertHooks orm.Hooks[[]Tset]
 
 	// NOTE: This is not called by InsertMany()
 	AfterInsertOneHooks orm.Hooks[T]
 	// NOTE: This is not called by UpsertMany()
 	AfterUpsertOneHooks orm.Hooks[T]
 
-	BeforeUpdateHooks orm.Hooks[T]
-	AfterUpdateHooks  orm.Hooks[T]
+	BeforeUpdateHooks orm.Hooks[Tslice]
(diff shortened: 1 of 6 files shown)


inoc603:
LGTM.

I was a little concerned about breaking backward compatibility, but it definitely feels better to not introducing new hooks.

stephenafamo:
> I was a little concerned about breaking backward compatibility, but it definitely feels better to not introducing new hooks.

Thankfully, Bob is still at `v0` so there is still room to break backward compatibility. However, I try not to only do it if absolutely necessary.

PR #81, opened 2023-07-18
Have separate update/delete queries for table models

stephenafamo:
This should address the issue in #75 

Let me know what you think @inoc603

### DIFF
--- a/clause/returning.go
+++ b/clause/returning.go
@@ -10,6 +10,10 @@ type Returning struct {
 	Expressions []any
 }
 
+func (r *Returning) HasReturning() bool {
+	return len(r.Expressions) > 0
+}
+
 func (r *Returning) AppendReturning(columns ...any) {
 	r.Expressions = append(r.Expressions, columns...)
 }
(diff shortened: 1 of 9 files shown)


inoc603:
So now if I were to use `UpdateAll`, looks like the entry would be:

```go
sqlResult, err := models.VideoTable.UpdateAll(ctx, db, mods...).Exec()
```

I don't have anything against it, but the `UpdateAll` method used to return `(int64, error)`. Maybe we could add a wrapper around it or return only `RowsAffected` so the return types can stay the same?

stephenafamo:
> Maybe we could add a wrapper around it or return only `RowsAffected` so the return types can stay the same?

Done. Makes sense to only return rows affected. Insert/Delete operations would not need `LastInsertID` anyway

inoc603:
LGTM. 

BTW Implementing bob.Mod on Setters is a really nice idea. It actually inspired me to make some of my own types into Mods and works really well.

stephenafamo:
Yeah. It makes me really like the query mod system.

PR #104, opened 2023-09-12
Implement named args and binding

stephenafamo:
Here is another attempt at binding named args.

Using it will look something like this:

```go
package main

import (
	"context"
	"fmt"

	"github.com/stephenafamo/bob"

	"github.com/stephenafamo/bob/dialect/psql"
	"github.com/stephenafamo/bob/dialect/psql/sm"
)

type Args struct {
	In1 int
	In2 int
	In3 int
	Id1 int
}

func main() {
	query := psql.Select(
		sm.Columns("id", "name"),
		sm.From("users"),
		sm.Where(psql.Quote("id").In(bob.Named("in1", "in2", "in3"))),
	)

	// Use as a regular query
	{
		bound, err := bob.BindNamed[*Args](query)
		if err != nil {
			panic(err)
		}

		queryStr, args, err := bound.Bind(nil).Build()
		if err != nil {
			panic(err)
		}

		fmt.Println(queryStr)
		fmt.Println(args)
	}

	// Use in prepared statements
	{
		ctx := context.Background()
		db, err := bob.Open("pgx", "postgres://postgres:postgres@localhost:5432/postgres?sslmode=disable")

		stmt, err := bob.Prepare[Args](ctx, db, query)
		if err != nil {
			panic(err)
		}

		_ = stmt
	}
}
```

@RangelReale let me know your thougts on this design.

1. The binding process can be used on any query, not just statements.
2. For a bob prepared statement, the `bob.InTx` method can be used to get a transaction specific version.

### DIFF
--- a/bound_query.go
+++ b/bound_query.go
@@ -6,10 +6,7 @@ import (
 	"reflect"
 
 	"github.com/stephenafamo/bob/internal/mappings"
-	"github.com/stephenafamo/scan"
 )
-
-var mapperSource, _ = scan.NewStructMapperSource()
 
 type MismatchedArgsError struct {
 	Expected int
--- a/bound_query.go
+++ b/bound_query.go
@@ -1,0 +1,169 @@
+package bob
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+
+	"github.com/stephenafamo/bob/internal/mappings"
+	"github.com/stephenafamo/scan"
+)
+
+var mapperSource, _ = scan.NewStructMapperSource()
+
+type MismatchedArgsError struct {
+	Expected int
+	Got      int
+}
+
+func (e MismatchedArgsError) Error() string {
+	return fmt.Sprintf("expected %d args, got %d", e.Expected, e.Got)
+}
+
+type DuplicateArgError struct{ Name string }
+
+func (e DuplicateArgError) Error() string {
+	return fmt.Sprintf("duplicate arg %s", e.Name)
+}
+
+type MissingArgError struct{ Name string }
+
+func (e MissingArgError) Error() string {
+	return fmt.Sprintf("missing arg %s", e.Name)
+}
+
+type binder[T any] interface {
+	ToArgs(T) []any
+}
+
+type BoundQuery[T any] struct {
+	query  []byte
+	binder binder[T]
+	start  int
+}
+
+func (b BoundQuery[T]) Bind(args T) BaseQuery[*cached] {
+	realArgs := b.binder.ToArgs(args)
+
+	return BaseQuery[*cached]{
+		Expression: &cached{
+			query: b.query,
+			args:  realArgs,
+			start: b.start,
+		},
+	}
+}
+
+func BindNamed[T any](q Query) (BoundQuery[T], error) {
+	return BindNamedN[T](q, 1)
+}
+
+func BindNamedN[T any](q Query, start int) (BoundQuery[T], error) {
+	query, args, err := BuildN(q, start)
+	if err != nil {
+		return BoundQuery[T]{}, err
+	}
+
+	binder, err := makeStructBinder[T](args)
+	if err != nil {
+		return BoundQuery[T]{}, err
+	}
+
+	return BoundQuery[T]{
+		query:  []byte(query),
+		binder: binder,
+		start:  start,
+	}, nil
+}
+
+type structBinder[T any] struct {
+	args   []string
+	fields []string
+}
+
+func (b structBinder[T]) Inspect() []string {
+	names := make([]string, len(b.args))
+	for _, name := range b.args {
+		if name == "" {
+			continue
+		}
+
+		names = append(names, name)
+	}
+
+	return names
+}
+
+func (b structBinder[T]) ToArgs(arg T) []any {
+	val := reflect.ValueOf(arg)
+	if val.Kind() == reflect.Pointer {
+		if val.IsNil() {
+			return make([]any, len(b.args))
+		}
+		val = val.Elem()
+	}
+
+	values := make([]any, len(b.args))
+
+ArgLoop:
+	for index, argName := range b.args {
+		for _, fieldName := range b.fields {
+			if fieldName == argName {
+				field := val.Field(index)
+				values[index] = field.Interface()
+				continue ArgLoop
+			}
+		}
+	}
+
+	return values
+}
+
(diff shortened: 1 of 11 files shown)


RangelReale:
Would it be possible for `BindNamed` to also accept a `map[string]any`?

RangelReale:
I did a first test with this design and it is very nice and clean, and this way I can use it with `sqlx` that we currently depend on 👍 
Will do a more in-depth test.

stephenafamo:
> Would it be possible for `BindNamed` to also accept a `map[string]any`?

It is easy to create a `mapBinder`, similar to the `structBinder`. As a matter of fact, I had the implementation ready but deleted it for now.

However, it would have to use its own entry function, e.g. `MapNamed`, and I left it this way for now to keep things easier to maintain.

RangelReale:
> Would it be possible for `BindNamed` to also accept a `map[string]any`?

The usecase is that I have some dynamic queries that can have more or less parameters depending on filters, otherwise I would need to create a struct with all possible parameters beforehand.

stephenafamo:
> I did a first test with this design and it is very nice and clean, and this way I can use it with `sqlx` that we currently depend on 👍
> Will do a more in-depth test

Thank you. I'm looking forward to the in-depth test to help iron out any issues with the design.

stephenafamo:
> The usecase is that I have some dynamic queries that can have more or less parameters depending on filters, otherwise I would need to create a struct with all possible parameters beforehand.

But how would this work since the actual query will expect the same number of `args` every time?
Or would these be args that can be `nil`?

RangelReale:
> > The usecase is that I have some dynamic queries that can have more or less parameters depending on filters, otherwise I would need to create a struct with all possible parameters beforehand.
> 
> But how would this work since the actual query will expect the same number of `args` every time? Or would these be args that can be `nil`?

For these queries I don't use `Prepare`, I build it every time, and add parameters using `Apply` and add parameters to the map.

```go
	if filter.Email != nil {
		query.Apply(
			sm.Where(`email = ?`, bob.Named("email")),
		)
		params["email"] = *filter.Email
	}
```

bound_query.go
```diff
@@ -0,0 +1,166 @@
...
+		Expression: &cached{
+			query: b.query,
+			args:  realArgs,
+			start: b.start,
+		},
+	}
+}
+
+func BindNamed[T any](q Query) (BoundQuery[T], error) {
+	return BindNamedN[T](q, 1)
```
RangelReale, line 55: Maybe I could have access the the generated SQL query by returning it from here or having an accessor method in `BoundQuery`? This way I could potentially use it to prepare the query in the DB, and use `Build` just to get the parameter replacement.

stephenafamo:
> For these queries I don't use `Prepare`, I build it every time, and add parameters using `Apply` and add parameters to the map.
> 
> ```go
> 	if filter.Email != nil {
> 		query.Apply(
> 			sm.Where(`email = ?`, bob.Named("email")),
> 		)
> 		params["email"] = *filter.Email
> 	}
> ```

If you're rebuilding the query every time, isn't it just easier to do this?
```go
 	if filter.Email != nil {
 		query.Apply(
 			sm.Where(psql.Quote("email").EQ(*filter.Email)),
 		)
 	}
 ```

This way you wouldn't even need to rebind the parameters.

File: bound_query.go (hunk 1 above)
Comment (stephenafamo) [reply bound_query.go:55]: If you need to prepare the query in the DB and bind as necessary, you should use the `bob.Prepare`/`bob.PrepareQuery` methods. Those have also been modified to take the `Arg` struct.

Or am I misunderstanding the use case?

**Comment by RangelReale** [comment]
> > For these queries I don't use `Prepare`, I build it every time, and add parameters using `Apply` and add parameters to the map.
> > ```go
> > 	if filter.Email != nil {
> > 		query.Apply(
> > 			sm.Where(`email = ?`, bob.Named("email")),
> > 		)
> > 		params["email"] = *filter.Email
> > 	}
> > ```
> 
> If you're rebuilding the query every time, isn't it just easier to do this?
> 
> ```go
>  	if filter.Email != nil {
>  		query.Apply(
>  			sm.Where(psql.Quote("email").EQ(*filter.Email)),
>  		)
>  	}
> ```
> 
> This way you wouldn't even need to rebind the parameters.

Will this generate and SQL argument or the text will be hardcoded on the SQL?

I tried to test it, but it seems that calling `BindNamed` expects that all arguments be `bob.Named`, this one which is not gave me an error:

```
panic: missing arg
```

**Comment by stephenafamo** [comment]
> Will this generate an SQL argument or the text will be hardcoded on the SQL?

> For these queries I don't use `Prepare`, I build it every time, and add parameters using `Apply` and add parameters to the map.
> 
> ```go
> 	if filter.Email != nil {
> 		query.Apply(
> 			sm.Where(`email = ?`, bob.Named("email")),
> 		)
> 		params["email"] = *filter.Email
> 	}
> ```

My bad 🤦🏾 . It should be like this to make it an arg.  (the old code wouldn't even compile).
```go
 	if filter.Email != nil {
 		query.Apply(
 			sm.Where(psql.Quote("email").EQ(psql.Arg(*filter.Email))),
 		)
 	}
 ```



> I tried to test it, but it seems that calling `BindNamed` expects that all arguments be `bob.Named`, this one which is not gave me an error:

It shouldn't. Any regular arg will be "saved" and returned in the final list after binding.

**Comment by stephenafamo** [comment]
> Seems to be not working like this:

Try now.

**Comment by RangelReale** [comment]
> > Seems to be not working like this:
> 
> Try now.

This seems fixed now, will resume my tests.

**Comment by RangelReale** [comment]
I'm finding that not having map parameters is somewhat limiting for the majority of queries, for example, `GetByID`:

```go
		query := psql.Select(
			sm.Columns(getDefaultTenantColumns()...),
			sm.From("tenant"),
			sm.Where(psql.Raw(`tenant_id = ?`, bob.Named("tenant_id"))),
			sm.Where(psql.Raw(`deleted_at IS NULL`)),
		)
```

In this case I would need to create a top level struct (as it needs to be declared on the type of `Stmt` stored on the `Storage` struct) just for this specific query, and I would need to create similar things for all queries expect the ones that uses all fields, like INSERT or UPDATE.

I don't thing arguments are reusable enough to require they always be a top-level declared struct, arguments vary a lot between queries.

stephenafamo:
@RangelReale after more than a year, I have returned to working on this 😅 

See details in #298

PR #105, opened 2023-09-12
Don't add parenthesis in RawQuery

RangelReale:
By [this change comment](https://github.com/stephenafamo/bob/pull/78#issuecomment-1715792764)

### DIFF
--- a/expr/raw.go
+++ b/expr/raw.go
@@ -20,6 +20,7 @@ func RawQuery(d bob.Dialect, q string, args ...any) bob.BaseQuery[Clause] {
 	return bob.BaseQuery[Clause]{
 		Expression: Clause{query: q, args: args},
 		Dialect:    d,
+		NoGroup:    true,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


stephenafamo:
While removing the parenthesis may help with readability, since the queries remain correct, I'll rather leave them in.

In many cases, Bob errs on the side of correctness over readability. While the `NoGroup` parameter is a small change, it will end up being something that has to be maintained indefinitely and edge cases may crop up.

Also, `RawQuery` is expected to be used for FULL queries e.g. (SELECT * FROM users). For raw query fragments, it is `psql.Raw()` that should be used. That does not add any parenthesis.

RangelReale:
`psql.Raw` also seems to add parenthesis:

```go
psql.Select(
  sm.Columns("id", "name"),
  sm.From("users"),
  sm.Where(psql.Raw(`id = 1`)),
)
```

```
SELECT 
id, name
FROM users
WHERE (id = 1)
```

Stepping in the code, it seems somehow this Where with Raw ends up in `expr.group.WriteSQL` which does end up adding parenthesis, I was not able to quickly find out where this `group` came from.

stephenafamo:
I'll take a look and figure out where it comes from.

RangelReale:
This also happens when using `psql.Raw` in Limit and Offset:

```go
		sm.Limit(psql.Raw(`?`, 10)),
		sm.Offset(psql.Raw(`?`, 11)),
```

```
LIMIT ($3)
OFFSET ($4)
```

this example don't make much sense in this way, but in my case I'm using `Raw` to set an sqlx named parameter. This probably will affect other uses of `Raw`.

PR #130, opened 2023-11-18
Generate setter if a table has relationships

stephenafamo:


aebranton:
Do we know when we can expect a release with this included? Seems to be work using existing release and just running the generation using @main since this is in it, but a release would be cool :)

stephenafamo:
A release is definitely due, although, I was hoping to take time out to fix one particular bug regarding multi sided relationships first.

aebranton:
No problem. Might be worth posting info that a workaround for non PK items is to run the generation using @main instead of a release tag for those that need it until a release is ready. 

Thanks for the library, breaking changes a couple times that have required resolving but overall I'm  a fan and use it in several projects

stephenafamo:
Released in `v0.23.0`

PR #153, opened 2024-01-15
Fix mysql alias in insert statement

jacobmolby:
MySQL cannot use aliasing when inserting, as far as I'm aware.

I get this error when running a 
```
models.Users.Insert(ctx,exec,setter)
```

```
 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'AS `users`
```

The generated sql looks something like:
```
INSERT INTO `users` AS `users`  (`id`, `created_at`, `updated_at`)
VALUES ((DEFAULT), (DEFAULT), (DEFAULT))
```
Previously Bob used `t.Name`. It was changed in this commit https://github.com/stephenafamo/bob/commit/5a2c154d851c59b2220487d9c862871c4ed6fe2b

### DIFF
--- a/dialect/mysql/table.go
+++ b/dialect/mysql/table.go
@@ -159,7 +159,7 @@ func (t *Table[T, Tslice, Tset]) InsertMany(ctx context.Context, exec bob.Execut
 	}
 
 	q := Insert(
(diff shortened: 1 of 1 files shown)


stephenafamo:
Thanks for fixing this. I somehow forgot that MySQL does not support this.

PR #156, opened 2024-01-16
Fix missing yaml struct tags

jacobmolby:
Playing around with 0.23.3 I've noticed that it didn't pick up my bobgen config with the new way of registering custom types.

If you have handled this in another branch, feel free to close this.


stephenafamo:
Oh, while you're at it, remove the `json` tags. That is what I wanted to delete, but deleted the wrong tag 🤦🏾 

Thanks for the PR

jacobmolby:
Done

PR #169, opened 2024-01-29
feat(bobgen-sqlite): support jsonb type as json

### LINKED ISSUE #168: JSONB support for SQLite
As of version 3.45.0, SQLite now supports `JSONB`. Thus, the generator should output it as `types.JSON[json.RawMessage]` like it does for `JSON`, instead of the default `string`.

hmnd:
fixes #168

### DIFF
--- a/gen/bobgen-sqlite/driver/sqlite.go
+++ b/gen/bobgen-sqlite/driver/sqlite.go
@@ -541,7 +541,7 @@ func (driver) translateColumnType(c drivers.Column) drivers.Column {
 		c.Type = "bool"
 	case "DATE", "DATETIME":
 		c.Type = "time.Time"
-	case "JSON":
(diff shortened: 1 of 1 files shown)


stephenafamo:
Thank you.

Kindly add a `JSONB` column to the sqlite driver test and make sure the type is set correctly in the golden files which would need to be updated by running the test with the flag `-overwrite-golden`.

Due to SQLite peculiarities, merely sending the JSON bytes to the column would not convert it into the JSONB without calling `jsonb()`.

So maybe not in this PR, but we should find a way for the generated models to always insert/update JSONB columns by wrapping the values in a jsonb function call.  
Or, if we can find a way to transform the bytes already, we can use a custom marshaller when sending the arguments (i.e. in `struct.Value()` method)

PR #184, opened 2024-02-20
config option to disable required rels for use in factories

jacobmolby:
Since we in our project don't use foreign keys, it can sometimes be annoying to work with the factories since they create relationships that hasn't been set.

This is a way to opt out of this behaviour.

It also allows to define relationships that might not always be true (for instance same id column used to reference different tables)

### DIFF
--- a/website/docs/code-generation/configuration.md
+++ b/website/docs/code-generation/configuration.md
@@ -197,6 +197,7 @@ We can manually describe relationships in the configuration:
 relationships:
   users: # The table name
     - name: "custom_videos_relationship" # A unique identifier used to configure aliases
+      no_foreign: true # If true, the factory will not generate the relationship, even if it's not set already.
       sides:
         - from: "users" # Name of the source of the relationship
           to: "videos" # Table name of the other side of the relation
--- a/website/docs/code-generation/configuration.md
+++ b/website/docs/code-generation/configuration.md
@@ -197,7 +197,7 @@ We can manually describe relationships in the configuration:
 relationships:
   users: # The table name
     - name: "custom_videos_relationship" # A unique identifier used to configure aliases
-      never_required: true # If true, the factory will not generate the relationship, even if it's not set already.
+      never_required: true # If true, the relationship would never be assumed to be required. This means that the factory will never auto-generate related objects for it.
       sides:
         - from: "users" # Name of the source of the relationship
           to: "videos" # Table name of the other side of the relation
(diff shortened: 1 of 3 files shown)


stephenafamo:
I don't think this is the correct way to do this.

When this configuration option is set, no relationship will be considered required even if it is based on a foreign key. This will make using the factory a pain.

I think instead, there should be an "identifier" on relationships to note that it is not based on a foreign key, and those should not be considered required (no flag needed)

jacobmolby:
I don't think I quite understand what you want. Do you want it to be configured via factory mod?

With how I made the PR, you have the option of turning it off or keeping it on if you want to preserve the default behavior, which I think is useful for many relationships.

Real foreign keys would be inferred right? So it wouldn't affect that (unless you manually override it in the config, which I don't think should be protected against.)

But again, I probably just don't understand what you wrote.

stephenafamo:
Maybe I'm the one missing something:


> Since we in our project don't use foreign keys, it can sometimes be annoying to work with the factories since they create relationships that hasn't been set.

If I understand correctly, this is possible because you define your relationships in the configuration correct?

If that is true, then my issue is that by having a `no-foreign` config option is too all-or-nothing. Users who mix auto-detected relationships and user-defined relationships would have to pick one of two extremes.

Instead, I want user-defined relationships to be defined and ignored in `relIsRequired`

stephenafamo:
> I don't really understand why you think this is an "all or nothing" approach since it would be defined on a relationship basis:

Oh! I misunderstood where the configuration was 🤦🏾 

I thought it was at the top level, similar to `no_factory`

stephenafamo:
However, I do not think `no_foreign` is an appropriate name, It does not properly convey what the option does.

Perhaps something like `never_required` would be more accurate. What do you think?

jacobmolby:
Yea, that's better :) I'll update the PR.

website/docs/code-generation/configuration.md
```diff
@@ -197,7 +197,7 @@ We can manually describe relationships in the configuration:
 relationships:
   users: # The table name
     - name: "custom_videos_relationship" # A unique identifier used to configure aliases
-      no_foreign: true # If true, the factory will not generate the relationship, even if it's not set already.
+      never_required: true # If true, the factory will not generate the relationship, even if it's not set already.
```
stephenafamo, line 200:
```suggestion
      never_required: true # If true, the relationship would never be assumed to be required. This means that the factory will never auto-generate related objects for it.
```

stephenafamo:
Thank you 🙏🏾

PR #186, opened 2024-02-26
Fixes for window functions

### LINKED ISSUE #185: [psql] Issues with window functions
The `window` argument to `Function.Over` is unused. 
https://github.com/stephenafamo/bob/blob/main/dialect/psql/dialect/function.go#L76

Additionally, `OVER ()` with an empty window definition is a valid statement, at least for PostgreSQL but is currently not possible due to `wr.def.Valid()` 
https://github.com/stephenafamo/bob/blob/main/dialect/psql/dialect/function.go#L122

relvacode:
Fixes #185

### DIFF
--- a/clause/window.go
+++ b/clause/window.go
@@ -23,26 +23,6 @@ type WindowDef struct {
 	Frame
 }
 
-func (wi WindowDef) Valid() bool {
-	if wi.From != "" {
-		return true
-	}
-
-	if len(wi.orderBy) != 0 {
-		return true
-	}
-
-	if len(wi.partitionBy) != 0 {
-		return true
-	}
-
-	if wi.Frame.Defined {
-		return true
-	}
-
-	return false
-}
-
 func (wi *WindowDef) SetFrom(from string) {
 	wi.From = from
 }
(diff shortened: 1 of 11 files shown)


stephenafamo:
Thank you so much for the PR. 

Can you add a test that uses a window function here?  
https://github.com/stephenafamo/bob/blob/main/dialect/psql/select_test.go

Also run `go generate` after to add it to the examples in the documentation.

relvacode:
Added, let me know if anything else needs changing

stephenafamo:
All looking good 👍🏾 

I also remembered by that argument was included, it was to support [this form of window functions `OVER window_name`](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-WINDOW-FUNCTIONS)

Would it be possible to support that with your changes?

stephenafamo:
Also, if `Valid()` is now unnecessary, kindly delete it.

relvacode:
Updated. Looks like `.From()` was already there to specify a named window.

stephenafamo:
Thanks for this PR.

I will later make some additional changes to properly construct `OVER window_name` window functions.

PR #188, opened 2024-03-04
Add support for []byte primary key types to gen

llamux:
Add type check during slice method generation to support []byte primary key type (BINARY in SQL land). Currently, this causes an error because comparison between []byte values is not supported using the equals operator. Using a BINARY(16) for the primary key is common in MySQL, since there is no native uuid type.


llamux:
After looking into it more, I'm doubtful this PR is worth it for the number of schemas affected.

stephenafamo:
Thanks for trying. I really appreciate it.

In this case, I would suggest replacing those columns with a dedicated UUID type using the [replacements configuration](https://bob.stephenafamo.com/docs/code-generation/configuration#replacements).  
This should also make working with it in code easier.

Issue #187, opened 2024-02-29
Code generateion not including the schema name


rkrishnasanka:
So for the following Atlas DDL,

```hcl

schema "core" {
  comment = "standard public schema"
}

table "tasks" {
    schema = schema.core
    comment = "Table for capturing operational data of task queue"

    column "id" {
        type = bigserial
        null = false
    }

    column "status" {
        type = enum.task_status_enum
        null = false
    }

    column "created_at" {
        type = timestamp
        null = false
        default = sql("CURRENT_TIMESTAMP")
    }

    column "completed_at" {
        type = timestamp
        null = true
    }

    column "payload" {
        // TODO- Change this to type = jsonb
        type = text
        null = true
    }

    column "queue_id" {
        type = bigint
        null = false
    }

    primary_key{
        columns = [column.id]
    }

    foreign_key "fk_task_queue_tasks_id" {
      columns = [column.queue_id]
      ref_columns = [table.queues.column.id]
    }
}

```

Bob is generating the following line in golang:

```go
// Tasks contains methods to work with the tasks table
var Tasks = psql.NewTablex[*Task, TaskSlice, *TaskSetter]("", "tasks")
```

it should generate the following line instead:

```go
// Tasks contains methods to work with the tasks table
var Tasks = psql.NewTablex[*Task, TaskSlice, *TaskSetter]("core", "tasks")
```

stephenafamo:
If you have only one schema, it is not included because it assumes that the schema would be included in the search path and the queries would be slightly less verbose.

If you want to force the addition of the schema name, set the `shared_schema` to another value. [see configuration](https://bob.stephenafamo.com/docs/code-generation/atlas#:~:text=shared_schema,first%20schema%20found)

rkrishnasanka:
Got it working @stephenafamo ! I don't think you should put the default value for `shared_schema` to pick the first schema from `schemas`. 

1. Its a non-obvious config problem that is very hard to diagnose when running the queries, half the time I wasn't sure if I configured something wrong or if my migration screwed something up.
2. Without the standard configuration example (full) its really hard to pinpoint this.

PR #189, opened 2024-03-04
Changed the default `shared_schema` options for the psql and atlas drivers

rkrishnasanka:
Updated the shared schema config to not default to the first schema. Verbose is better. Currently only updated the psql and atlas drivers. Related to the conversation in #187

### DIFF
--- a/gen/bobgen-atlas/driver/atlas.go
+++ b/gen/bobgen-atlas/driver/atlas.go
@@ -119,10 +119,6 @@ func (d *driver) Assemble(ctx context.Context) (*DBInfo, error) {
 		return nil, err
 	}
 
-	if d.config.SharedSchema == "" {
-		d.config.SharedSchema = realm.Schemas[0].Name
-	}
-
 	d.loadEnums(realm)
 	dbinfo = &DBInfo{
 		Enums:  d.getEnums(),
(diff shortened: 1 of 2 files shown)


rkrishnasanka:
@stephenafamo let me know if this makes sense. It'll definitely solve my be issue of an assumption that you can easily miss in the config

stephenafamo:
> It'll definitely solve my be issue of an assumption that you can easily miss in the config

I'm hesitant to change this.

Most people would only be working with one schema, and may not appreciate the extra verbosity of including the schema name when they are only working with one schema.

And while I understand that they could manually set the shared schema to get the same behaviour back, I prefer to have the "default behaviour" be as seamless as possible.

If the issue is that is is unclear what shared schema is being used, wouldn't it be more beneficial to add a log statement to make the user aware of this?

What do you think?

rkrishnasanka:
Well I guess its a reasonable alternative. I'm personally a fan of explicit specifications rather than navigating default behavior, but I understand where you might be coming from. However, you should give it some thought, having explicit verbose configs are the easiest ways of learning how processes work. 

For all the people who try to deploy the tool for anything beyond the basic usage, not having explicit complete configurations is a pain to deal with. For me, it changed what I could have done in 1 hour to stretch over a week as I was trying to figure out why things weren't working.

PR #191, opened 2024-03-11
Update to koanf v2

mmrath:
Update https://github.com/knadh/koanf to v2


stephenafamo:
Thanks for this.

But shouldn't the import path for the providers also be updated to `github.com/knadh/koanf/v2/providers/...`?

mmrath:
Hey @stephenafamo no the import paths for provides are not under `github.com/knadh/koanf/v2` they are now separate modules, previously they were part of `github.com/knadh/koanf`

stephenafamo:
Alright.

PR #212, opened 2024-05-17
Fix multi sided relationships attachment

jacobmolby:
When having a multisided relationship like this:

```yml
 users:
    - name: "users_to_teams_through_team_users"
      sides:
        - from: "users"
          to: "team_user"
          columns:
          - [id,user_id]
        - from: "team_user"
          to: "teams"
          columns:
          - [team_id,id]

```

With the following tables:
```sql
CREATE TABLE `users` (
  `id` int unsigned NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`id`),
) ENGINE=InnoDB AUTO_INCREMENT=12 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;

CREATE TABLE `teams` (
  `id` bigint unsigned NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`id`),
) ENGINE=InnoDB AUTO_INCREMENT=7 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;

CREATE TABLE `team_user` (
  `id` bigint unsigned NOT NULL AUTO_INCREMENT,
  `team_id` int unsigned NOT NULL,
  `user_id` int unsigned NOT NULL,
  PRIMARY KEY (`id`),
  KEY `team_user_team_id_user_id_index` (`team_id`,`user_id`)
) ENGINE=InnoDB AUTO_INCREMENT=7 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

It produces the wrong sql for attaching `users` to `teams` and vice-versa. Notice that the `TeamID` is missing, which results in it not being attached properly.

`models/teams.go`
```golang
func attachTeamUsers0(ctx context.Context, exec bob.Executor, count int, teamUsers1 TeamUserSlice, team0 *Team, users2 UserSlice) (TeamUserSlice, error) {
	for i := range teamUsers1 {
		setter := &TeamUserSetter{
			UserID: omit.From(users2[i].ID),
		}

		err := TeamUsers.Update(ctx, exec, setter, teamUsers1[i])
		if err != nil {
			return nil, fmt.Errorf("attachTeamUsers0, update %d: %w", i, err)
		}
	}

	return teamUsers1, nil
}
```

This change makes sure the `team0` variable is used, so the output becomes the following:

```golang
func attachTeamUsers0(ctx context.Context, exec bob.Executor, count int, teamUsers1 TeamUserSlice, team0 *Team, users2 UserSlice) (TeamUserSlice, error) {
	for i := range teamUsers1 {
		setter := &TeamUserSetter{
			TeamID: omit.From(team0.ID),
			UserID: omit.From(users2[i].ID),
		}

		err := TeamUsers.Update(ctx, exec, setter, teamUsers1[i])
		if err != nil {
			return nil, fmt.Errorf("attachTeamUsers0, update %d: %w", i, err)
		}
	}

	return teamUsers1, nil
}
```

I don't know if the changes I made are optimal or (even wrong for some) however for my code base it produced the correct output.

### DIFF
--- a/gen/templates/models/11_rel_ops.go.tpl
+++ b/gen/templates/models/11_rel_ops.go.tpl
@@ -113,9 +113,9 @@
     for i := range {{$sideAlias.DownPlural}}{{$side.Position}} {
       setter := &{{$sideAlias.UpSingular}}Setter{
         {{range $map := $side.Mapped -}}
+          {{$colName := $sideAlias.Column $map.Column -}}
           {{if $rel.NeedsMany .ExtPosition -}}
             {{$sideC := $sideTable.GetColumn .Column -}}
-            {{$colName := $sideAlias.Column $map.Column -}}
             {{if .HasValue -}}
               {{if $sideC.Nullable }}
                 {{$.Importer.Import "github.com/aarondl/opt/omitnull" -}}
@@ -146,6 +146,10 @@
                 {{$colName}}: omit.From({{$colVal}}),
               {{- end}}
             {{- end}}
+          {{- else}}
+            {{$a := $.Aliases.Table .ExternalTable -}}
+            {{$colVal := printf "%s%d.%s" $a.DownSingular $map.ExtPosition ($a.Column $map.ExternalColumn) -}}
+            {{$colName}}: omit.From({{$colVal}}), 
           {{- end}}
         {{- end}}
       }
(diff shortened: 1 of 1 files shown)


stephenafamo:
All the checks pass, so I assume it's all good too. Thanks for the contribution

stephenafamo:
Can you edit a few things?

1. Make sure that we're not creating more variables than necessary
2. Possibly rename the variables in the template to be a bit more descriptive.  
  As you know, the template can very quickly become unreadable so these things help 🙏🏾

PR #234, opened 2024-06-08
fix psql inet type without subnet

gstarikov:
Hi. 
Postgres inet type may contain /32 or /128 subnet and in that cases postgress will return string without '/' and netip.ParsePrefix will fail with error.

### DIFF
--- a/gen/bobgen-helpers/helpers.go
+++ b/gen/bobgen-helpers/helpers.go
@@ -219,11 +219,11 @@ func Types() drivers.Types {
                 return any(types.Text[netip.Addr, *netip.Addr]{Val: ipAddr}).(T)`,
 			RandomExprImports: importers.List{`"crypto/rand"`},
 		},
-		"postgres.Inet": {
+		"pgtypes.Inet": {
 			Imports: importers.List{
 				`"net/netip"`,
 				`"strings"`,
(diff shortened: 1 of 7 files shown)


stephenafamo:
However, I'm currently working on this, alongside other changes to types as a whole, so this is badly timed 😬 

I'll close this PR now, but please send in other PRs if you have the time. I really appreciate the help 🙏🏾

stephenafamo:
Actually, this has already been released in `v0.27.0`

<https://github.com/stephenafamo/bob/releases/tag/v0.27.0#:~:text=Use%20netip.Addr%20instead%20of%20netip.Prefix%20for%20Postgres%20cidr%20type>

gstarikov:
there is support for cidr postgres type in `0.27.0` , but i need to fix inet support. in case of inet type and single ip subnet (/32 or /128) postgres will return only ip addres, without '/' suffix and netip.ParsePrefix will fail with error.

stephenafamo:
I understand now. This is very helpful.

stephenafamo (changes requested):
Kindly move the new type from `types/postgres/inet.go` to `types/pgtypes/inet.go`

I think that is a more appropriate package name (also that is where I'm placing some new types which I am creating).

Issue #120, opened 2023-11-02
How to access nonunique indexes for tables in code generation?


majiaxin110:
I use this library extensively to generate an orm that is somewhat customized. I wonder if there is a way to access and output all the index names of a table when I write a custom template for code generation.

Accessing indexes masters especially when index hints must be provided. (https://dev.mysql.com/doc/refman/8.0/en/index-hints.html). 

I find there are primary keys, foreign keys, and constraints in gen/drivers/Table. It seems nonunique indexes are missed.

For now, I may choose to implement a TemplateDataPlugin, which accesses all indexes and writes them to ExtraInfo in TemplateData. Is it a good practice? Or may I use such feature in the future?

Thanks a lot!

stephenafamo:
Up till now, retrieving index information was not a priority so it was not included.

It would of course be a welcome addition, so if you can send in a PR for this (even if just for MySQL), I would appreciate it.

stephenafamo:
Fixed in #237

PR #237, opened 2024-06-16
Load index information for MySQL, PostgreSQL, and SQLite tables

mbezhanov:
Hi, I am opening this in relation to https://github.com/stephenafamo/bob/issues/120.

This will allow users like @majiaxin110 to range over `.Table.Indexes` in their custom `*.tpl` enabling the use of `$index.Name` variables in a similar fashion:

```
{{$table := .Table}}
{{$tAlias := .Aliases.Table $table.Key -}}
{{range $index := .Table.Indexes}}
func print_index_{{$tAlias.DownSingular}}_{{$index.Name}}() {
    fmt.Println("{{$index.Name}}")
}
{{end}}
```

Sample output:

```go
func print_index_jet_idx_jets_names() {
	fmt.Println("idx_jets_names")
}

func print_index_jet_idx_jets_pilots() {
	fmt.Println("idx_jets_pilots")
}

func print_index_jet_PRIMARY() {
	fmt.Println("PRIMARY")
}
```

I only added MySQL support to keep the PR focused and get some early feedback on the implementation. If you approve of it, I'd happily add support for PostgreSQL and SQLite as well.

### DIFF
--- a/gen/drivers/indexes.go
+++ b/gen/drivers/indexes.go
@@ -1,0 +1,7 @@
+package drivers
+
+// DBIndexes lists all indexes in the database schema keyed by table name
+type DBIndexes map[string][]Index
+
+// Index represents an index in a table
+type Index NamedColumnList
--- a/gen/bobgen-atlas/driver/atlas.mysql_golden.json
+++ b/gen/bobgen-atlas/driver/atlas.mysql_golden.json
@@ -83,6 +83,7 @@
 					"type": "int32"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_multi_keys",
@@ -138,6 +139,7 @@
 					"type": "int32"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_sponsors",
@@ -166,6 +168,7 @@
 					"type": "int32"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_tags",
@@ -1536,6 +1539,7 @@
 					"type": "string"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_type_monsters",
@@ -1586,6 +1590,7 @@
 					"type": "int32"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": null,
 				"foreign": null,
@@ -1609,6 +1614,7 @@
 					"type": "int32"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_users",
@@ -1648,6 +1654,7 @@
 					"type": "int32"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_video_tags",
@@ -1720,6 +1727,7 @@
 					"type": "int32"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_videos",
--- a/gen/bobgen-atlas/driver/atlas.psql_golden.json
+++ b/gen/bobgen-atlas/driver/atlas.psql_golden.json
@@ -17,6 +17,7 @@
 					"type": "int32"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_sponsors",
@@ -45,6 +46,7 @@
 					"type": "int32"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_tags",
@@ -1833,6 +1835,7 @@
 					"type": "string"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_type_monsters",
@@ -1883,6 +1886,7 @@
 					"type": "string"
 				}
 			],
+			"indexes": null,
 			"constraints": {
 				"primary": {
 					"name": "pk_users",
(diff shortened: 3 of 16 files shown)


stephenafamo:
This is very nice! Exactly what was requested in #120

I'll take a deeper look, but overall looks great 👌🏾

mbezhanov:
Happy to hear that! I also included PostgreSQL support.

stephenafamo:
The tests are failing because the new output does not match the golden files....

You should run the generation tests locally with the `-overwrite-golden` flag to fix this.

```sh
go test ./gen/bobgen-psql/driver -overwrite-golden
```

gen/drivers/indexes.go
```diff
@@ -0,0 +1,10 @@
+package drivers
+
+// DBIndexes lists all indexes in the database schema keyed by table name
+type DBIndexes map[string][]Index
+
+// Index represents an index in a table
+type Index struct {
+	Name    string   `yaml:"name" json:"name"`
+	Columns []string `yaml:"columns" json:"columns"`
+}
```
stephenafamo, line 10: We should instead reuse the `Constraint` type since they have the same structure.

Perhaps it makes more sense for the type to be called `Index` or something else, but I would rather no duplication.

mbezhanov: Even though both types have the same structure, an `Index` and a `Constraint` have different semantics. Indexes are automatically created for primary key columns and unique constraints, but things are different for foreign keys. MySQL (InnoDB) automatically creates indexes for them, but PostgreSQL and SQLite do not.

Take the following DDL statements, for example, executed against a PostgreSQL database:

```sql
CREATE TABLE test.parent (
   id INT GENERATED ALWAYS AS IDENTITY,
   name VARCHAR(255) NOT NULL,
   PRIMARY KEY(id)
);

CREATE TABLE test.child (
   id INT GENERATED ALWAYS AS IDENTITY,
   parent_id INT,
   name VARCHAR(255) NOT NULL,
   PRIMARY KEY(id),
   CONSTRAINT fk_parent FOREIGN KEY(parent_id) REFERENCES parent(id)
);
```

If we run the following query to list their indexes:

```sql
SELECT * FROM pg_catalog.pg_indexes WHERE schemaname = 'test';
```

We get the following result:

```text
schemaname|tablename|indexname  |tablespace|indexdef                                                       |
----------+---------+-----------+----------+---------------------------------------------------------------+
test      |parent   |parent_pkey|          |CREATE UNIQUE INDEX parent_pkey ON test.parent USING btree (id)|
test      |child    |child_pkey |          |CREATE UNIQUE INDEX child_pkey ON test.child USING btree (id)  |
```

I can certainly generalize the struct to avoid code duplication, but neither `Index` nor `Constraint` will be a semantically correct name.

stephenafamo: Yes, I understand that semantically, they are different things. This is why I said perhaps it should be called something different (not Index or Constraint).

It just feels a bit redundant to have two types that are exactly the same thing.

I guess a bit of repetition can be good for the sake of correctness.

stephenafamo:
Can you also update the SQLite golden file?

gen/drivers/indexes.go
mbezhanov: What do you think about something similar:

```go
type NamedColumnList struct {
	Name    string   `yaml:"name" json:"name"`
	Columns []string `yaml:"columns" json:"columns"`
}


type Constraint struct {
	NamedColumnList
}

type Index struct {
	NamedColumnList
}
```

Or would you rather completely eliminate the `Index` type?

mbezhanov: This will allow the `Index` and `Constraint` types to evolve separately (if needed).

stephenafamo: > This will allow the `Index` and `Constraint` types to evolve separately (if needed).

If that time comes, then we can create these wrapper types. In the meantime, we can replace every instance with `Constraint` with `NamedColumnList`.

If we really want to have distinct types now, we can do 

```go
type Constraint NamedColumnList

type Index NamedColumnList
```

This is also great since there are no methods attached to `NamedColumnList`

mbezhanov:
> Can you also update the SQLite golden file?

Yes, absolutely. I'll actually go ahead and directly implement the SQLite index retrieval logic in this PR. Coming up in a little bit :slightly_smiling_face: 

Would you like me to squash everything up to this point into a single commit, or are you planning to squash when merging into main?

stephenafamo:
> Would you like me to squash everything up to this point into a single commit, or are you planning to squash when merging into main?

Kindly squash and push.

I do not squash when merging into main, so the commit history can get a little messy if it is not done beforehand 😬

mbezhanov:
Pushing an updated version rebased on the latest `main`. I've added index listing support to `bobgen-mysql`, `bobgen-psql`, and `bobgen-sqlite`. I can do `bobgen-atlas` and `bobgen-prisma` in a separate PR, unless you prefer to have them included in this one.

I introduced a `NamedColumnList` struct as discussed, but kept `Index` and `Constraint` as distinct types rather than aliases for better readability:

```go
type Constraint NamedColumnList

type Index NamedColumnList
```

I found the aliases a bit confusing during debug sessions:

image

Type declarations seem to be a bit more clear:

image

However, if you insist on type aliases rather than type declarations, I'd be happy to change them.

Let me know if any further improvements are necessary.

stephenafamo:
Looks great! Thank you.

stephenafamo:
I've already merged this now, but I just thought that it would be nice to be able to tell apart different types of indexes (unique vs non-unique)

mbezhanov:
No problem. I'll add this and open another PR in the following days.

PR #239, opened 2024-06-18
Load index information from Atlas schema files

mbezhanov:
Continuing what I started in #237, this PR adds proper support for loading index information from Atlas schema files.

I wanted to do the Atlas and Prisma driver modifications together, but the Atlas implementation took a bit longer than I anticipated.

I had to update the original `schema.hcl` files, as they had some small issues preventing the `atlas` CLI from executing them against real MySQL, PostgreSQL, and SQLite databases (now it works).

All questions and comments are welcome.

### DIFF
--- a/gen/bobgen-atlas/driver/atlas.go
+++ b/gen/bobgen-atlas/driver/atlas.go
@@ -174,6 +174,7 @@ func (d *driver) tables(realm *schema.Realm) []drivers.Table {
 					Foreign: fks,
 				},
 				Columns: d.tableColumns(atlasTable, colFilter),
+				Indexes: d.indexes(atlasTable, colFilter),
 			}
 			tables = append(tables, table)
 		}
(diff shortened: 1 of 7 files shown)


gen/bobgen-atlas/driver/atlas.go
```diff
@@ -540,3 +541,69 @@ func (p *driver) getEnums() []drivers.Enum {
...
+				case "sqlite":
+					isCompositePrimaryKey := len(table.PrimaryKey.Parts) > 1
+					isIntegerColumnType := false
+					typ, ok := table.PrimaryKey.Parts[0].C.Type.Type.(*schema.IntegerType)
+					if ok {
+						isIntegerColumnType = typ.T == "integer"
+					}
+					if isCompositePrimaryKey || !isIntegerColumnType {
+						pkName = "sqlite_autoindex_" + table.Name + "_1"
+					}
```
mbezhanov, line 578: When the following conditions are met, SQLite doesn't create an `sqlite_autoindex` on the provided primary key:

```c
if( nTerm==1
 && pCol
 && pCol->eCType==COLTYPE_INTEGER
 && sortOrder!=SQLITE_SO_DESC
){
```

The fragment above is from the SQLite source code:

* `nTerm` is the number of columns supplied to the `PRIMARY KEY`
* `pCol` is the first column
* `pCol->eCType` is the column type. SQLite insists on this being declared as an `integer` and not an `int`. If the column is defined as an `int` then an index will be created.
* `sortOrder` is not really applicable to our use case, as I don't think the Atlas schema allows you to define a `PRIMARY KEY DESC` column (but the SQLite specification does).

stephenafamo: Can you share a link to this? 🙏🏾 


### Side note

A part of me is considering dropping the Atlas and Prisma drivers. 
Since they can be used to do migrations, perhaps the better flow is `prisma/atlas -> migrate -> bob gen from db`

I am not sure how many people are using these drivers, but I'm conflicted about the cost of maintaining them and if it is worth it.

stephenafamo: I think I will take them out.

At best, I could keep them alive in a separate repository, but I'm less and less convinced that it should be part of the core.

Kindly hold off on the Prisma implementation. You already did a lot with this one. I really appreciate it.

mbezhanov: Okay, I'll just fix the linting errors on this one then, and let you decide whether to merge or discard it.

stephenafamo: Thanks you once more for your amazing contribution 🙏🏾

Issue #244, opened 2024-06-26
Cannot load index information for functional indexes


pcriv:
After upgrading to `0.28.0` from `0.27.1`, I'm getting the following error when trying to generate code from my schema:

```
unable to fetch table data: unable to load indexes: sql: Scan error on column index 2, name "column_name": converting NULL to string is unsupported
```

stephenafamo:
Can you share more details?

* What driver are you using?
* What does your config look like?
* An example schema to reproduce this

pcriv:
driver: mysql
config:

```yaml
---
mysql:
  dsn: "root:@tcp(localhost:3306)/dname"
  output: internal/dbgen
  pkgname: dbgen
  except:
    schema_migrations:
relationships:
  # hidden
```

About the schema, I cannot share the whole schema, but is there any specific part that you have in mind that could help?

stephenafamo:
> is there any specific part that you have in mind that could help?

The part that helps to reproduce the error.

Perhaps you can use the `only` configuration to generate table-by-table till you find the offending table and share a schema of that table?

pcriv:
That's a good idea! I'll try that and keep you updated, thanks!

pcriv:
I think I found the offensive line.
FYI, the `only` option did not work, it was failing no matter which table I added, so I had to get creative

This is a bit of an ugly one:
```
KEY `index_batches_tags_workspace_id` (((cast(json_unquote(json_extract(`tags`,_utf8mb4'$.workspaceId')) as char(255) charset utf8mb4) collate utf8mb4_bin)))
```

stephenafamo:
This probably has something to do with #237 

Can you perhaps take a look @mbezhanov?

mbezhanov:
Sure, I'll have a look right away.

mbezhanov:
Okay, I've identified the issue. This is a functional index (it indexes expressions rather than columns), so `column_names` in `information_schema.statistics` is indeed `NULL`. I'm working on a fix.

pcriv:
Thanks for the quick reactions @mbezhanov and @stephenafamo ❤️

mbezhanov:
No problem at all, and sorry for my oversight!

PR #246, opened 2024-06-27
Handle `null` column names in expression indexes

mbezhanov:
This provides a fix to #244.

DDL for testing:

```sql
CREATE TABLE t1 (
  col1 int,
  col2 int,
  col3 int
);

CREATE INDEX idx1 ON t1 ((col1 + col2));
CREATE INDEX idx2 ON t1 ((col1 + col2), col3);
CREATE INDEX idx3 ON t1 (col3);
```

Before the patch, generation failed with the following error:

1. MySQL:

    > 2024/06/27 11:49:12 unable to fetch table data: unable to load indexes: sql: Scan error on column index 2, name "column_name": converting NULL to string is unsupported

2. SQLite:
    > 2024/06/27 11:51:14 unable to fetch table data: sql: Scan error on column index 0, name "name": converting NULL to string is unsupported

After the patch, generation completes successfully with the following information being collected:

image

---

Please note that PostgreSQL was not affected by this issue, as it uses the expression as a `column_name` rather than omitting it:

image

---

@stephenafamo would you prefer to use the expression contents as a `column_name` for SQLite and MySQL as well, rather than omitting them? I checked two different database GUIs. The first one used the expression contents as a `column_name`, and the others one omitted the expression from the column listing.

### DIFF
--- a/gen/bobgen-atlas/driver/atlas.mysql_golden.json
+++ b/gen/bobgen-atlas/driver/atlas.mysql_golden.json
@@ -88,27 +88,31 @@
 					"name": "PRIMARY",
 					"columns": [
 						"id"
-					]
+					],
+					"expressions": null
 				},
 				{
 					"name": "one",
 					"columns": [
 						"one",
 						"two"
-					]
+					],
+					"expressions": null
 				},
 				{
 					"name": "something",
 					"columns": [
 						"something",
 						"another"
-					]
+					],
+					"expressions": null
 				},
 				{
 					"name": "sponsor_id",
 					"columns": [
 						"sponsor_id"
-					]
+					],
+					"expressions": null
 				}
 			],
 			"constraints": {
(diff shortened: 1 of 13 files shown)


stephenafamo:
I think it would be better to have a different field in the struct to contain the expression.

This way, someone could generate code to use the expression in a query.

mbezhanov:
I've updated the PR accordingly.

How do you feel about this format for the `Index` structures:

image

That corresponds to the following DDL:

```sql
CREATE TABLE test_index_expressions (
    col1 int,
    col2 int,
    col3 int
);
CREATE INDEX idx11 ON test_index_expressions (col1, (col2 + col3), (POW(col3, 2)))
```

stephenafamo:
This looks good 🎉 

Perhaps it makes sense to have a flat struct for `Index`

```go
type Index struct {
  Name string
  Columns []string
  Expressions []string
}
```

mbezhanov:
> Perhaps it makes sense to have a flat struct for `Index`

:white_check_mark: Done

PR #248, opened 2024-07-10
Export custom unique constraint error types

mbezhanov:
Opening this PR to materialize our discussion from #242.

In its current form, it adds support for the following drivers via `bobgen.yaml`:

```yaml
psql:
  driver_name: "pq" # default
#  driver_name: "pgx"
#  driver_name: "pgx/v4"
#  driver_name: "pgx/v5"

sqlite:
  driver_name: "sqlite" # default
#  driver_name: "go-sqlite3"
```

The MySQL driver is non-configurable and just defaults to `github.com/go-sql-driver/mysql`.

There's one bigger design decision that's worth discussing before finalizing this PR.

Right now, I'm generalizing the main `errUniqueConstraint` structure in `bob_main.go` as follows:
```go
type errUniqueConstraint struct {
	// s is a string uniquely identifying the constraint in 
        // the raw error message returned from the database.
	s string
}
```

Here, `s` is a substring to look for in the raw error message that the specified database system has returned. The reason why I'm using this generalization is because Postgres and MySQL report the constraint name, but SQLite reports the names of the columns included in the constraint (omitting the name).

This is also why in `01_types.go.tpl`, I'm doing this to build the respective substring:

```go
{{ $s := $constraint.Name }}
{{ if eq "sqlite" $.Dialect }}
{{ $s = printf "%s.%s" $table.Name (join (printf ", %s." $table.Name) $constraint.Columns) }}
{{ end }}
```

A possible alternative would be to keep the `errUniqueConstraint` type more verbose, e.g.:

```go
type errUniqueConstraint struct {
	name string
	table  string
	cols[] string
}
```

And then just use the `table` and `cols` information to augment the `unique_constraint_error_detection_method` block for SQLite to produce something like:

```go
func (e *errUniqueConstraint) Is(target error) bool {
	err, ok := target.(*sqliteDriver.Error)
	if !ok {
		return false
	}
	var sb strings.Builder
	for i := 0; i < len(e.cols); i++ {
		if i > 0 {
			sb.WriteString(", ")
		}
		sb.WriteString(e.table + "." + e.cols[i])
	}
	return err.Code() == 2067 && strings.Contains(err.Error(), sb.String())
}
```

Once we come to an agreement on that decision (and address any other comments that may arise), I'll finalize this PR with additional tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,9 +41,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Added custom error types to generated models for checking unique constraint errors.
+
 ### Fixed
 
 - Remove unnecessary import of `strings` in `bobfactory_random.go`.
+- Prevent data races in unit tests.
 
 ## [v0.28.1] - 2024-06-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,12 +43,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added custom error types to generated models for checking unique constraint errors.
+- Added error constants for matching against both specific and generic unique constraint errors raised by the underlying database driver.
 
 ### Fixed
 
-- Remove unnecessary import of `strings` in `bobfactory_random.go`.
-- Prevent data races in unit tests.
+- Removed unnecessary import of `strings` in `bobfactory_random.go`.
+- Fixed data races in unit tests.
 
 ## [v0.28.1] - 2024-06-28
 
--- a/gen/bobgen-atlas/driver/atlas.mysql_golden.json
+++ b/gen/bobgen-atlas/driver/atlas.mysql_golden.json
@@ -1899,5 +1899,6 @@
 			]
 		}
 	],
-	"extra_info": null
+	"extra_info": null,
+	"driver_name": ""
 }
--- a/gen/bobgen-atlas/driver/atlas.psql_golden.json
+++ b/gen/bobgen-atlas/driver/atlas.psql_golden.json
@@ -2122,5 +2122,6 @@
 			]
 		}
 	],
-	"extra_info": null
+	"extra_info": null,
+	"driver_name": ""
 }
--- a/gen/bobgen-atlas/driver/atlas.sqlite_golden.json
+++ b/gen/bobgen-atlas/driver/atlas.sqlite_golden.json
@@ -1765,5 +1765,6 @@
 		}
 	],
 	"enums": [],
-	"extra_info": null
+	"extra_info": null,
+	"driver_name": ""
 }
--- a/gen/bobgen-mysql/driver/mysql.go
+++ b/gen/bobgen-mysql/driver/mysql.go
@@ -98,7 +98,7 @@ func (d *driver) Assemble(ctx context.Context) (*DBInfo, error) {
 	}
 	defer d.conn.Close()
 
-	dbinfo = &DBInfo{}
+	dbinfo = &DBInfo{DriverName: "github.com/go-sql-driver/mysql"}
 
 	dbinfo.Tables, err = drivers.BuildDBInfo(ctx, d, d.config.Concurrency, d.config.Only, d.config.Except)
 	if err != nil {
--- a/gen/bobgen-mysql/driver/mysql.golden.json
+++ b/gen/bobgen-mysql/driver/mysql.golden.json
@@ -1993,5 +1993,6 @@
 			]
 		}
 	],
-	"extra_info": null
+	"extra_info": null,
+	"driver_name": "github.com/go-sql-driver/mysql"
 }
--- a/gen/bobgen-mysql/templates/models/singleton/bob_mysql_blocks.go.tpl
+++ b/gen/bobgen-mysql/templates/models/singleton/bob_mysql_blocks.go.tpl
@@ -6,3 +6,15 @@ var (
 	DeleteWhere = Where[*dialect.DeleteQuery]()
 )
 {{- end -}}
+
+{{define "unique_constraint_error_detection_method"}}
+{{$.Importer.Import "strings"}}
+{{$.Importer.Import "mysqlDriver" "github.com/go-sql-driver/mysql"}}
+func (e *errUniqueConstraint) Is(target error) bool {
+  err, ok := target.(*mysqlDriver.MySQLError)
+  if !ok {
+    return false
+  }
+  return err.Number == 1062 && strings.Contains(err.Message, e.s)
+}
+{{end}}
--- a/gen/bobgen-prisma/driver/prisma.mysql_golden.json
+++ b/gen/bobgen-prisma/driver/prisma.mysql_golden.json
@@ -1736,5 +1736,6 @@
 			"DriverSource": "",
 			"DriverENVSource": ""
 		}
-	}
+	},
+	"driver_name": ""
 }
--- a/gen/bobgen-prisma/driver/prisma.psql_golden.json
+++ b/gen/bobgen-prisma/driver/prisma.psql_golden.json
@@ -1736,5 +1736,6 @@
 			"DriverSource": "",
 			"DriverENVSource": ""
 		}
-	}
+	},
+	"driver_name": ""
 }
--- a/gen/bobgen-prisma/driver/prisma.sqlite_golden.json
+++ b/gen/bobgen-prisma/driver/prisma.sqlite_golden.json
@@ -1736,5 +1736,6 @@
 			"DriverSource": "",
 			"DriverENVSource": ""
 		}
-	}
+	},
+	"driver_name": ""
 }
--- a/gen/bobgen-psql/driver/psql.go
+++ b/gen/bobgen-psql/driver/psql.go
@@ -47,6 +47,8 @@ type Config struct {
 	Concurrency int
 	// Which UUID package to use (gofrs or google)
 	UUIDPkg string `yaml:"uuid_pkg"`
+	// Which `database/sql` driver to use (the full module name)
+	DriverName string `yaml:"driver_name"`
 
 	//-------
 
@@ -69,6 +71,10 @@ func New(config Config) Interface {
 
 	if config.UUIDPkg == "" {
 		config.UUIDPkg = "gofrs"
+	}
+
+	if config.DriverName == "" {
+		config.DriverName = "github.com/lib/pq"
 	}
 
 	if config.Concurrency < 1 {
@@ -132,7 +138,7 @@ func (d *driver) Assemble(ctx context.Context) (*DBInfo, error) {
 	}
 	defer d.conn.Close()
 
-	dbinfo = &DBInfo{}
+	dbinfo = &DBInfo{DriverName: d.config.DriverName}
 
 	// drivers.Tables call translateColumnType which uses Enums
 	if err := d.loadEnums(ctx); err != nil {
--- a/gen/bobgen-psql/driver/psql.golden.json
+++ b/gen/bobgen-psql/driver/psql.golden.json
@@ -4135,5 +4135,6 @@
 			]
 		}
 	],
-	"extra_info": null
+	"extra_info": null,
+	"driver_name": "github.com/lib/pq"
 }
--- a/gen/bobgen-psql/main.go
+++ b/gen/bobgen-psql/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"io/fs"
 	"log"
 	"os"
 	"os/signal"
@@ -49,7 +50,10 @@ func run(c *cli.Context) error {
 	}
 
 	d := driver.New(driverConfig)
-	outputs := helpers.DefaultOutputs(driverConfig.Output, driverConfig.Pkgname, config.NoFactory, nil)
+	outputs := helpers.DefaultOutputs(
+		driverConfig.Output, driverConfig.Pkgname, config.NoFactory,
+		&helpers.Templates{Models: []fs.FS{gen.PSQLModelTemplates}},
+	)
 
 	state := &gen.State{
 		Config:  config,
--- a/gen/bobgen-psql/templates/models/singleton/bob_psql_blocks.go.tpl
+++ b/gen/bobgen-psql/templates/models/singleton/bob_psql_blocks.go.tpl
@@ -25,7 +25,7 @@ func (e *errUniqueConstraint) Is(target error) bool {
 	if !ok {
 		return false
 	}
-	return err.Code == "23505" && err.{{$constraintNameField}} == e.s
+	return err.Code == "23505" && (e.s == "" || err.{{$constraintNameField}} == e.s)
 	{{end}}
 }
 {{end}}
--- a/gen/bobgen-psql/templates/models/singleton/bob_psql_blocks.go.tpl
+++ b/gen/bobgen-psql/templates/models/singleton/bob_psql_blocks.go.tpl
@@ -1,0 +1,31 @@
+{{define "unique_constraint_error_detection_method"}}
+func (e *errUniqueConstraint) Is(target error) bool {
+	{{$supportedDrivers := list "github.com/lib/pq" "github.com/jackc/pgx" "github.com/jackc/pgx/v4" "github.com/jackc/pgx/v5"}}
+	{{if not (has $.DriverName $supportedDrivers)}}
+	return false
+	{{else}}
+		{{$errType := ""}}
+		{{$constraintNameField := "ConstraintName"}}
+		{{if eq $.DriverName "github.com/jackc/pgx/v4" "github.com/jackc/pgx/v5"}}
+			{{$errType = "*pgconn.PgError"}}
+			{{if eq $.DriverName "github.com/jackc/pgx/v4" }}
+				{{$.Importer.Import "github.com/jackc/pgconn"}}
+			{{else}}
+				{{$.Importer.Import "github.com/jackc/pgx/v5/pgconn"}}
+			{{end}}
+		{{else}}
+			{{$.Importer.Import $.DriverName}}
+			{{$errType = "pgx.PgError"}}
+			{{if eq $.DriverName "github.com/lib/pq" }}
+				{{$errType = "*pq.Error"}}
+				{{$constraintNameField = "Constraint"}}
+			{{end}}
+		{{end}}
+	err, ok := target.({{$errType}})
+	if !ok {
+		return false
+	}
+	return err.Code == "23505" && err.{{$constraintNameField}} == e.s
+	{{end}}
+}
+{{end}}
--- a/gen/bobgen-sqlite/driver/sqlite.go
+++ b/gen/bobgen-sqlite/driver/sqlite.go
@@ -22,6 +22,9 @@ type (
 )
 
 func New(config Config) Interface {
+	if config.DriverName == "" {
+		config.DriverName = "modernc.org/sqlite"
+	}
 	return &driver{config: config}
 }
 
@@ -39,6 +42,8 @@ type Config struct {
 	Only map[string][]string
 	// List of tables that will be should be ignored. Others are included
 	Except map[string][]string
+	// Which `database/sql` driver to use (the full module name)
+	DriverName string `yaml:"driver_name"`
 
 	// Used in main.go
 
@@ -106,7 +111,12 @@ func (d *driver) Assemble(ctx context.Context) (*DBInfo, error) {
 		return nil, err
 	}
 
-	return &DBInfo{Tables: tables}, nil
+	dbinfo := &DBInfo{
+		DriverName: d.config.DriverName,
+		Tables:     tables,
+	}
+
+	return dbinfo, nil
 }
 
 func (d *driver) buildQuery(schema string) (string, []any) {
@@ -428,7 +438,7 @@ func (d driver) foreignKeys(ctx context.Context, schema, tableName string) ([]dr
 	return fkeys, nil
 }
 
-// uniques retrieves the foreign keys for a given table name.
+// uniques retrieves the unique keys for a given table name.
 func (d driver) uniques(ctx context.Context, schema, tableName string) ([]drivers.Constraint, error) {
 	rows, err := d.conn.QueryContext(ctx, fmt.Sprintf("PRAGMA '%s'.index_list('%s')", schema, tableName))
 	if err != nil {
--- a/gen/bobgen-sqlite/driver/sqlite.golden.json
+++ b/gen/bobgen-sqlite/driver/sqlite.golden.json
@@ -2396,5 +2396,6 @@
 		}
 	],
 	"enums": null,
-	"extra_info": null
+	"extra_info": null,
+	"driver_name": "modernc.org/sqlite"
 }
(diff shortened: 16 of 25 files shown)


stephenafamo:
Thank you for this!

> In its current form, it adds support for the following drivers via `bobgen.yaml`:
> 
> ```yaml
> psql:
>   driver_name: "pq" # default
> #  driver_name: "pgx"
> #  driver_name: "pgx/v4"
> #  driver_name: "pgx/v5"
> 
> sqlite:
>   driver_name: "sqlite" # default
> #  driver_name: "go-sqlite3"
> ```
> 
> The MySQL driver is non-configurable and just defaults to `github.com/go-sql-driver/mysql`.

I'd rather use the full import path as the driver name. Best for correctness and prevent having to jump hoops in case we run into similar named drivers down the line.

Even currently with SQLite, from first look it is hard to tell which driver configuration corresponds to which driver vs seeing `github.com/mattn/go-sqlite`


> I'm generalizing the main `errUniqueConstraint` structure in `bob_main.go`

This is fine. For the same end result, it is best to generate less code

stephenafamo:
> ```go
> return err.Code() == 2067 && strings.Contains(err.Error(), sb.String())
> ```

Do we need to do string matching? I believe these drivers (at least the postgres drivers) should directly include the constraint name:

* `lib/pq`: <https://pkg.go.dev/github.com/lib/pq#Error:~:text=string%0A%09DataTypeName%20%20%20%20%20string-,Constraint,-string%0A%09File%20%20%20%20%20%20%20%20%20%20%20%20%20string>
* `pgx/v5`: <https://pkg.go.dev/github.com/jackc/pgx/v5@v5.6.0/pgconn#PgError:~:text=string%0A%09DataTypeName%20%20%20%20%20%20%20%20string-,ConstraintName,-string%0A%09File%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20string>

mbezhanov:
> Do we need to do string matching? I believe these drivers (at least the postgres drivers) should directly include the constraint name.

Thanks for pointing this out. I have reworked the implementation accordingly.

Couldn't do it for the MySQL and SQLite drivers, though, as their error types don't contain any fields exposing only the constraint names.

mbezhanov:
I've updated the tests as well, and everything seems to be green now, so I'm marking this PR as ready for review.

Is there a recommended way to provide automated tests verifying the generated `func (e *errUniqueConstraint) Is(target error) { ... }` body for each distinct driver? (I haven't written any).

I haven't worked on adapting the Atlas and Prisma drivers to these changes as well. Because of that, I had to add this extra conditional check to the `bob_psql_blocks.go.tpl` and `bob_sqlite_blocks.go.tpl` files to make tests pass for Atlas and Prisma:

```go
{{if or (not $.ExtraInfo) (not (hasField $.ExtraInfo "DriverName"))}}
```

Not sure if you're okay with that or not, so let me know if you prefer something else.

stephenafamo:
> Is there a recommended way to provide automated tests verifying the generated

No recommended way for now, there are the tests generated for `ScannerValuer` and others for the random function. Maybe you can do something similar.

 

> I haven't worked on adapting the Atlas and Prisma drivers to these changes as well. Because of that, I had to add this extra conditional check to the `bob_psql_blocks.go.tpl` and `bob_sqlite_blocks.go.tpl` files to make tests pass for Atlas and Prisma:

Since all 3 major drivers have the `driverName` field, perhaps it is better to move it into the main DBInfo type instead of using `Extra` for it.

mbezhanov:
I've updated the code, promoting `DriverName` to a top-level `DBInfo` field.

As for the tests, I tried a few different approaches, but they all end up gravitating towards generating integration tests in either `bob_main_test.go.tpl` or maybe in a newly introduced `01_types_test.go.tpl` file, so I'm not sure I'm on the right track here.

These tests would basically insert a duplicate record into a test database table, and then verify that the error can be checked using `errors.Is()` and the newly introduced `errUniqueConstraint` type.

Is this idea worth exploring further?

stephenafamo:
> As for the tests, I tried a few different approaches, but they all end up gravitating towards generating integration tests in either `bob_main_test.go.tpl` or maybe in a newly introduced `01_types_test.go.tpl` file, so I'm not sure I'm on the right track here.

You're on the right track. I think `01_types_test.go.tpl` would be the way to go.

> Is this idea worth exploring further?

Yes..... However, from my experience with SQLBoiler, the issue is that it could easily break.

The clean way would be to use the factory to generate a random model and insert it twice. This breaks because a user can have constraints on the table that Bob is unaware of.

To make this usable by everyone, there should be a way to hook in the user's default factory mods in these tests.

mbezhanov:
I put a `*.tpl` together, capable of dynamically generating the following output:

<details>
   <summary><b>pilots_test.go</b></summary>

```go
package test

import (
  "bobtest/models"
  "bobtest/models/factory"
  "context"
  "database/sql"
  "errors"
  "testing"

  _ "github.com/go-sql-driver/mysql"
  "github.com/stephenafamo/bob"
)

func TestPilotUniqueConstraintErrors(t *testing.T) {
  dsn := "" // TODO: fill
  db, err := sql.Open("mysql", dsn)
  if err != nil {
    t.Fatal("Error connecting to database")
  }
  tests := []struct {
    name        string
    expectedErr error
    applyFn     func(tpl *factory.PilotTemplate, obj *models.Pilot)
  }{
    {name: "ErrUniqueFirstNameAndLastName", expectedErr: models.PilotErrors.ErrUniqueFirstNameAndLastName, applyFn: func(tpl *factory.PilotTemplate, obj *models.Pilot) {
      tpl.Apply(
        factory.PilotMods.FirstName(obj.FirstName),
        factory.PilotMods.LastName(obj.LastName),
      )
    }},
  }

  for _, tt := range tests {
    t.Run(tt.name, func(t *testing.T) {
      ctx := context.Background()
      exec := bob.New(db)
      f := factory.New()
      tpl := f.NewPilot()
      obj, err := tpl.Create(ctx, exec)
      if err != nil {
        t.Fatal(err)
      }
      tt.applyFn(tpl, obj)
      _, err = tpl.Create(ctx, exec)
      if !errors.Is(tt.expectedErr, err) {
        t.Fatalf("Expected %s, Got: %s", tt.name, err.Error())
      }
    })
  }
}
```

</details>

It can switch between different drivers and everything. However, there are two problems:

1. Test files have to be placed outside the `models` package. Otherwise, you get a circular dependency between `models` and `models/factory`. This results in the following directory structure, which I'm not sure is okay:
    image
2. It's unclear how to automatically populate the test DSN. Therefore, it's up to the user to fill it in, which I'm not sure is okay:
    ```go
      dsn := "" // TODO: fill
      db, err := sql.Open("mysql", dsn)
      ```

Other than that, the tests seem to work fine, even though I haven't tested edge cases too extensively.

If you're happy with this, I can commit for you to review. If not, let me know what further changes would be necessary.

stephenafamo:
> 1. Test files have to be placed outside the `models` package. Otherwise, you get a circular dependency between `models` and `models/factory`.

Will using a test package solve this? e.g.:

```go
// models/tablename_test.go
package models_test
```

> 2\. It's unclear how to automatically populate the test DSN.

We should use an environment variable. Not sure what convention to go with.. maybe `PSQL_TEST_DSN`?

mbezhanov:
I updated the PR, taking into account what we discussed, and tested everything a bit more thoroughly. This results in similar test files being generated for each model:

<details><summary><b>orders_test.go</b></summary>

```go
// Code generated by BobGen mysql (devel). DO NOT EDIT.
// This file is meant to be re-generated in place and/or deleted at any time.

package models_test

import (
  "context"
  "database/sql"
  "errors"
  "os"
  "testing"

  models "bobtest/models"
  factory "bobtest/models/factory"
  _ "github.com/go-sql-driver/mysql"
  "github.com/stephenafamo/bob"
)

func TestOrderUniqueConstraintErrors(t *testing.T) {
  db, err := sql.Open("mysql", os.Getenv("MYSQL_TEST_DSN"))
  if err != nil {
    t.Fatal("Error connecting to database")
  }
  tests := []struct {
    name        string
    expectedErr error
    applyFn     func(tpl *factory.OrderTemplate, obj *models.Order)
  }{
    {
      name:        "ErrUniqueOrderNumber",
      expectedErr: models.OrderErrors.ErrUniqueOrderNumber,
      applyFn: func(tpl *factory.OrderTemplate, obj *models.Order) {
        tpl.Apply(
          factory.OrderMods.RandomizeAllColumns(nil),
          factory.OrderMods.OrderNumber(obj.OrderNumber),
        )
      },
    },
    {
      name:        "ErrUniqueUserIdAndOrderDate",
      expectedErr: models.OrderErrors.ErrUniqueUserIdAndOrderDate,
      applyFn: func(tpl *factory.OrderTemplate, obj *models.Order) {
        tpl.Apply(
          factory.OrderMods.RandomizeAllColumns(nil),
          factory.OrderMods.UserID(obj.UserID),
          factory.OrderMods.OrderDate(obj.OrderDate),
        )
      },
    },
  }

  for _, tt := range tests {
    t.Run(tt.name, func(t *testing.T) {
      ctx := context.Background()
      tx, err := db.Begin()
      if err != nil {
        t.Fatal("Couldn't start database transaction")
      }
      exec := bob.New(tx)
      f := factory.New()
      tpl := f.NewOrder()
      obj, err := tpl.Create(ctx, exec)
      if err != nil {
        t.Fatal(err)
      }
      tt.applyFn(tpl, obj)
      _, err = models.Orders.Insert(ctx, exec, tpl.BuildSetter())
      if !errors.Is(tt.expectedErr, err) {
        t.Fatalf("Expected: %s, Got: %v", tt.name, err)
      }
      if err = tx.Rollback(); err != nil {
        t.Fatal("Couldn't rollback database transaction")
      }
    })
  }
}
```

</details>

Generation was performed against the following database schemas:

<details><summary><b>MySQL</b></summary>

```sql
--- MySQL
CREATE TABLE users (
    id INT AUTO_INCREMENT PRIMARY KEY,
    first_name VARCHAR(50) NOT NULL,
    last_name VARCHAR(50) NOT NULL,
    email VARCHAR(100) NOT NULL,
    UNIQUE (email),
    UNIQUE (first_name, last_name)
);

CREATE TABLE orders (
    id INT AUTO_INCREMENT PRIMARY KEY,
    user_id INT NOT NULL,
    order_number VARCHAR(50) NOT NULL,
    order_date DATE NOT NULL,
    UNIQUE (order_number),
    UNIQUE (user_id, order_date),
    FOREIGN KEY (user_id) REFERENCES users(id)
);
```

</details>

<details><summary><b>Postgres</b></summary>

```sql
--- Postgres
CREATE TABLE users (
    id SERIAL PRIMARY KEY,
    first_name VARCHAR(50) NOT NULL,
    last_name VARCHAR(50) NOT NULL,
    email VARCHAR(50) NOT NULL,
    UNIQUE (email),
    UNIQUE (first_name, last_name)
);

CREATE TABLE orders (
    id SERIAL PRIMARY KEY,
    user_id INT NOT NULL,
    order_number VARCHAR(50) NOT NULL,
    order_date DATE NOT NULL,
    UNIQUE (order_number),
    UNIQUE (user_id, order_date),
    FOREIGN KEY (user_id) REFERENCES users(id)
);
```

</details>


<details><summary><b>SQLite</b></summary>

```sql
--- SQLite
CREATE TABLE users (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    first_name TEXT NOT NULL,
    last_name TEXT NOT NULL,
    email TEXT NOT NULL,
    UNIQUE (email),
    UNIQUE (first_name, last_name)
);

CREATE TABLE orders (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    user_id INTEGER NOT NULL,
    order_number TEXT NOT NULL,
    order_date TEXT NOT NULL,
    UNIQUE (order_number),
    UNIQUE (user_id, order_date),
    FOREIGN KEY (user_id) REFERENCES users(id)
);
```

</details>

Using the following commands:

```bash
# MySQL
export MYSQL_TEST_DSN=test:test@/second?parseTime=true && go test ./models -v

# Postgres
export PSQL_TEST_DSN=postgres://postgres:test@localhost/second?sslmode=disable && go test ./models -v

# SQLite
export SQLITE_TEST_DSN=/home/marin/Projects/GolandProjects/bob/file.db && go test ./models -v
```

Expected results:

```bash
=== RUN   TestOrderUniqueConstraintErrors
=== RUN   TestOrderUniqueConstraintErrors/ErrUniqueOrderNumber
=== RUN   TestOrderUniqueConstraintErrors/ErrUniqueUserIdAndOrderDate
--- PASS: TestOrderUniqueConstraintErrors (0.01s)
    --- PASS: TestOrderUniqueConstraintErrors/ErrUniqueOrderNumber (0.01s)
    --- PASS: TestOrderUniqueConstraintErrors/ErrUniqueUserIdAndOrderDate (0.00s)
=== RUN   TestUserUniqueConstraintErrors
=== RUN   TestUserUniqueConstraintErrors/ErrUniqueEmail
=== RUN   TestUserUniqueConstraintErrors/ErrUniqueFirstNameAndLastName
--- PASS: TestUserUniqueConstraintErrors (0.00s)
    --- PASS: TestUserUniqueConstraintErrors/ErrUniqueEmail (0.00s)
    --- PASS: TestUserUniqueConstraintErrors/ErrUniqueFirstNameAndLastName (0.00s)
PASS
ok      bobtest/models  0.019s
```

* Database transactions are used to prevent the test data from polluting the database.
* No test methods are generated if `no_factory: true` in `bobgen.yaml`.
* No test methods are generated if there aren't any unique contraints to check in the given table.

This hasn't been tested against unique constraints over computed fields (i.e., unique indexes using functions/expressions) and likely won't work properly with them, but is a good starting point for providing such support in a subsequent PR.

mbezhanov:
@stephenafamo I am not sure why the PR tests aren't passing, as everything is green locally.

```bash
export MYSQL_DRIVER_TEST_DSN=test:test@/testdroppable?multiStatements=true && go test ./...

?       github.com/stephenafamo/bob/clause      [no test files]
?       github.com/stephenafamo/bob/dialect/mssql       [no test files]
ok      github.com/stephenafamo/bob     0.002s
?       github.com/stephenafamo/bob/dialect/mysql/dialect       [no test files]
?       github.com/stephenafamo/bob/dialect/mysql/dm    [no test files]
?       github.com/stephenafamo/bob/dialect/mysql/fm    [no test files]
?       github.com/stephenafamo/bob/dialect/mysql/im    [no test files]
?       github.com/stephenafamo/bob/dialect/mysql/sm    [no test files]
?       github.com/stephenafamo/bob/dialect/mysql/um    [no test files]
?       github.com/stephenafamo/bob/dialect/psql/dialect        [no test files]
?       github.com/stephenafamo/bob/dialect/psql/dm     [no test files]
?       github.com/stephenafamo/bob/dialect/psql/fm     [no test files]
?       github.com/stephenafamo/bob/dialect/psql/im     [no test files]
?       github.com/stephenafamo/bob/dialect/psql/sm     [no test files]
?       github.com/stephenafamo/bob/dialect/psql/um     [no test files]
?       github.com/stephenafamo/bob/dialect/sqlite/dialect      [no test files]
?       github.com/stephenafamo/bob/dialect/sqlite/fm   [no test files]
?       github.com/stephenafamo/bob/dialect/sqlite/sm   [no test files]
?       github.com/stephenafamo/bob/dialect/sqlite/um   [no test files]
?       github.com/stephenafamo/bob/dialect/sqlite/dm   [no test files]
?       github.com/stephenafamo/bob/dialect/sqlite/im   [no test files]
?       github.com/stephenafamo/bob/gen/bobgen-atlas    [no test files]
ok      github.com/stephenafamo/bob/dialect/mysql       1.271s
?       github.com/stephenafamo/bob/gen/bobgen-helpers  [no test files]
?       github.com/stephenafamo/bob/gen/bobgen-prisma   [no test files]
?       github.com/stephenafamo/bob/gen/bobgen-mysql    [no test files]
?       github.com/stephenafamo/bob/gen/bobgen-psql     [no test files]
?       github.com/stephenafamo/bob/gen/bobgen-sql      [no test files]
?       github.com/stephenafamo/bob/gen/bobgen-sqlite   [no test files]
?       github.com/stephenafamo/bob/internal/mappings   [no test files]
?       github.com/stephenafamo/bob/models      [no test files]
?       github.com/stephenafamo/bob/models/factory      [no test files]
?       github.com/stephenafamo/bob/mods        [no test files]
?       github.com/stephenafamo/bob/test/files  [no test files]
?       github.com/stephenafamo/bob/test/utils  [no test files]
?       github.com/stephenafamo/bob/types       [no test files]
?       github.com/stephenafamo/bob/types/pgtypes       [no test files]
?       github.com/stephenafamo/bob/website     [no test files]
ok      github.com/stephenafamo/bob/dialect/psql        5.689s
ok      github.com/stephenafamo/bob/dialect/sqlite      0.167s
ok      github.com/stephenafamo/bob/expr        0.006s
ok      github.com/stephenafamo/bob/gen 0.004s
ok      github.com/stephenafamo/bob/gen/bobgen-atlas/driver     135.004s
ok      github.com/stephenafamo/bob/gen/bobgen-mysql/driver     62.873s
ok      github.com/stephenafamo/bob/gen/bobgen-prisma/driver    123.309s
ok      github.com/stephenafamo/bob/gen/bobgen-psql/driver      96.225s
ok      github.com/stephenafamo/bob/gen/bobgen-sql/driver       95.992s
ok      github.com/stephenafamo/bob/gen/bobgen-sqlite/driver    86.002s
ok      github.com/stephenafamo/bob/gen/drivers 0.021s
ok      github.com/stephenafamo/bob/gen/importers       0.383s
ok      github.com/stephenafamo/bob/internal    0.007s
ok      github.com/stephenafamo/bob/orm 0.011s
```

stephenafamo:
Amazing work. Thank you!

I'll take a look at the failing tests.

mbezhanov:
After a bit of error and trial, I was able to reproduce it locally:

```sh
--- FAIL: TestPostgres (69.15s)
    --- FAIL: TestPostgres/generate (2.41s)
        gen.go:137: Unable to execute State.Run: unable to generate output: failed to format template
            
              20        "github.com/stephenafamo/bob/mods"
              21        "github.com/stephenafamo/bob/orm"
              22 )
              23 
              24 
            >>>>                VideoID int32 `db:"video_id,pk" `
              26                TagID int32 `db:"tag_id,pk" `
              27 
              28        R videoTagR `db:"-" `
              29        }
              30 
            
            :25:3: expected declaration, found VideoID
        testing.go:1398: race detected during execution of test
```

Seems like a data race error (`templateByteBuffer` getting corrupted due to concurrency). Every time I re-run the test suite, it occurs on random files and locations.

Currently looking into it, but saw this in `output.go`:

```go
. . .

var (
  // templateByteBuffer is re-used by all template construction to avoid
  // allocating more memory than is needed. This will later be a problem for
  // concurrency, address it then.
  templateByteBuffer       = &bytes.Buffer{}
  templateHeaderByteBuffer = &bytes.Buffer{}

. . . 
```

Maybe you already have an idea what needs to change?

mbezhanov:
I provided a fix in https://github.com/stephenafamo/bob/pull/248/commits/eca1ef661226b3d6c2be5a076fb15e778372853e. Let me know if that's not what you had in mind, though, and if any changes might be needed.

stephenafamo:
Thanks for the amazing work.

I've been having an extremely busy couple of weeks, so I've not had time to take a proper look.  
I'll try to go through all of this and merge it before next weekend.

You should also add your changes to `CHANGELOG.md` under `Unreleased` 🙏🏾

mbezhanov:
I've updated the changelog as requested. I won't be available next week, but if anything else has to be addressed, I'll be happy to do it after that.

stephenafamo:
> I think it would be helpful to also export the generic one that wraps the specific ones.

100% this.

We shouldn't break any code which looks out for the original error.

mbezhanov:
Do you have any preferences for what this should look like?

Because the easiest change would to make `errUniqueContraint` an exported type, and then users could utilize it like this:

```go
if errors.Is(&models.ErrUniqueConstraint{}, err) {
  log.Fatal(err)
}
```

Alternatively, I can export a generic type for each model, and then users will be able to utilize it like this:

```go
if errors.Is(models.PilotErrors.ErrUniqueConstraint, err) {
  log.Fatal(err)
}
```

stephenafamo:
I misread @daddz comment. I thought by the "original error" they were referring to the error returned from the database itself.  
But they are of course referring to just the new type. And not exporting that would not break any code.

I also don't fully understand the exact benefit of exporting

> handle unique constraint error in general

@daddz to handle unique constraint errors in general, the user should follow the instructions of the appropriate driver without needing the generated code.  
Is there a reason why Bob needs to help out with this?

stephenafamo:
@mbezhanov Can you please add a bit of documentation about this feature on this page?  
https://bob.stephenafamo.com/docs/code-generation/usage

mbezhanov:
@stephenafamo I added some documentation. Tried to keep it short and concise, following the style of the preceding documentation. Let me know if you need more details though.

@daddz Just to clarify, I am not doing anything with the raw DB driver error right now (i.e., I am not wrapping it or anything). Rather, I am introducing a custom error type that implements an `Is()` method, which allows me to take advantage of the following Go errors package feature:

> An error type might provide an Is method so it can be treated as equivalent to an existing error.

So the raw database error is stored in the `err` variable, and `errors.Is()` uses the specific unique constraint error (e.g., `ErrUniqueFirstNameAndLastName`) to check for equivalence:

```go
pilot, err := models.Pilots.Insert(ctx, db, setter)
if errors.Is(models.PilotErrors.ErrUniqueFirstNameAndLastName, err) {
	log.Fatal(err)
}
```

`ErrUniqueFirstNameAndLastName`, however, does not wrap the raw `err` in any way. It also doesn't wrap some generic unique constraint error.

The generic check might be implemented in a similar way:

```go
var PilotErrors = &pilotErrors{
	// ErrUniqueConstraint captures all unique constraint errors by leaving `s` empty.
	ErrUniqueConstraint:           &errUniqueConstraint{s: ""},
	ErrUniqueFirstNameAndLastName: &errUniqueConstraint{s: "pilots_first_name_last_name_key"},
}

type pilotErrors struct {
	ErrUniqueConstraint           error
	ErrUniqueFirstNameAndLastName error
}
```

mbezhanov:
@stephenafamo @daddz I've pushed the generic unique constraint error implementation for you to review. I'll revert it if it turns out to be unfit for release.

mbezhanov:
Thanks for the clarification. In that case, I'll have to rework my last commit if we are to expose a single generic `models.ErrUniqueConstraint`. It can certainly be done, though, if that's the preferred way of using it.

mbezhanov:
I reworked the implementation accordingly. Users can now do:

```go
if !errors.Is(models.PilotErrors.ErrUniqueFirstNameAndLastName, err) {
  // do something specific
}

if !errors.Is(models.ErrUniqueConstraint, err) {
  // do something generic
}
```

stephenafamo:
Looks great.

Kindly also update the docs/changelog to note the generic error

mbezhanov:
> Kindly also update the docs/changelog to note the generic error

:heavy_check_mark: Done

Rupikz:
@stephenafamo Can you please add a bit of documentation about new config parameters "driver_name" on this page https://bob.stephenafamo.com/docs/code-generation/psql and for other drivers?

stephenafamo:
I would appreciate if you could send in a PR 🙏🏾

mbezhanov:
@Rupikz  @stephenafamo Hope this helps: https://github.com/stephenafamo/bob/pull/342

 If you need additional details on how `driver_name` works, just let me know and I'll add more info.

PR #265, opened 2024-08-08
Add note for disabling SSL mode in PSQL DSN if needed.

Hasaber8:
wrt SSL connections, you will get an error 
`⁠ unable to fetch table data: unable to load enums: pq: SSL is not enabled on the server ⁠`

so I’m adding notes to docs - that you can let pq know SSL is disabled.


Hasaber8:
Screenshot 2024-08-08 at 23 12 45
Page preview.

PR #268, opened 2024-08-16
Fix invalid SQL generated by OrderBy().Collate()

mbezhanov:
Fixes #252

Tested with the following schemas and code:

<details>
  <summary><b>MySQL</b></summary>

  ```sql
  CREATE TABLE users (
    id INT AUTO_INCREMENT PRIMARY KEY,
    name VARCHAR(255)
  );
  INSERT INTO users (name) VALUES ('béatrice'), ('bérénice'), ('bernard'), ('boris');
  SELECT * FROM users ORDER BY name COLLATE utf8mb4_bin ASC;
  SELECT * FROM users ORDER BY name COLLATE utf8mb4_unicode_ci ASC;
  ```

  ```go
  ctx := context.Background()
  
  ex, err := bob.Open("mysql", "test:test@/test?parseTime=true")
  if err != nil {
    log.Fatal(err)
  }
  
  type userObj struct {
    Id   int
    Name string
  }
  
  cls := []string{"utf8mb4_bin", "utf8mb4_unicode_ci"}
  
  for _, cl := range cls {
    q := mysql.Select(
      sm.Columns("id", "name"),
      sm.From("users"),
      sm.OrderBy(models.UserColumns.Name).Collate(cl).Asc(),
    )
  
    users, err := bob.All(ctx, ex, q, scan.StructMapper[userObj]())
    if err != nil {
      log.Fatal(err)
    }
  
    fmt.Println()
    fmt.Println(cl)
    for _, u := range users {
      fmt.Println(u)
    }
  }
  ```
</details>

<details>
  <summary><b>PostgreSQL</b></summary>

  ```sql
  CREATE TABLE users (
    id SERIAL PRIMARY KEY,
    name TEXT
  );
  INSERT INTO users (name) VALUES ('béatrice'), ('bérénice'), ('bernard'), ('boris');
  SELECT * FROM users ORDER BY name COLLATE "C" asc;
  SELECT * FROM users ORDER BY name COLLATE "fr-FR-x-icu";
  ```

  ```go
  ctx := context.Background()
  
  ex, err := bob.Open("postgres", "postgres://postgres:test@localhost/test?sslmode=disable")
  if err != nil {
    log.Fatal(err)
  }
  
  type userObj struct {
    Id   int
    Name string
  }
  
  cls := []string{"C", "fr-FR-x-icu"} 
  
  for _, cl := range cls {
    q := psql.Select(
      sm.Columns("id", "name"),
      sm.From("users"),
[...]

PR #270, opened 2024-08-17
Remove MS SQL artifacts

mbezhanov:
Opening this to facilitate a discussion. Please feel free to close this PR if the removals it proposes seem unnecessary.

I noticed these two files while working on #268. If I understand correctly, MS SQL support was initially planned but eventually abandoned.

Can you confirm that these files are no longer needed for the project? If they're not, I feel it may be good to remove them to avoid confusion.


stephenafamo:
I previously left them since I felt it wasn't affecting anything else.  
However, you're correct. It is better to remove it until support for SQL Server is actually being worked on.

Issue #293, opened 2024-11-02
Include CASE WHEN expression starter


TimurUtyaganov:
I'm looking for a convenient way to use case when clause

stephenafamo:
Can you give an example of the query you're trying to build?

TimurUtyaganov:
> Can you give an example of the query you're trying to build?

select b.id, b.created_at from blocks b where case when b.type = 'periodic' then b.created_at - interval '10' day <= now() when b.type = 'test' then b.color in (...) end. 
Something like that. Thx for replying

stephenafamo:
No expression helper for this yet.... I'll add this later

k4n4ry:
@stephenafamo 
Hello, I'd like to help resolve this issue. Would you mind if I work on it?

stephenafamo:
Sure. Go ahead.

My suggestion:

1. Since the `CASE` syntax is the same across all 3 dialects (Postgres, MySQL and SQLite), create a CASE expr in the `expr` package and then use it in `dialect/starters.go`
2. the base expression itself should return a "Chainable" mod that can be used to add `WHEN` and `END` clauses. 
    ```go
    expr.Case().                                             // should not yet be a valid expression
        When(condition bob.Expression, then bob.Expression). // can stop here and be a valid expression
        When(condition bob.Expression, then bob.Expression). // can stop here and be a valid expression
        Else(bob.Expression)                                 // can stop here and be valid
    ```
    Look for `JoinChain` for an example of something chainable, although that is used to build a `bob.Mod`

PR #294, opened 2024-11-06
Include CASE WHEN expression starter

k4n4ry:
Implements #293

### DIFF
--- a/dialect/mysql/starters.go
+++ b/dialect/mysql/starters.go
@@ -1,9 +1,6 @@
 package mysql
 
 import (
-	"context"
-	"io"
-
 	"github.com/stephenafamo/bob"
 	"github.com/stephenafamo/bob/dialect/mysql/dialect"
 	"github.com/stephenafamo/bob/expr"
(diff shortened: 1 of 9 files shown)


expr/case.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"context"
+	"errors"
+	"io"
+
+	"github.com/stephenafamo/bob"
+)
+
+type Case[T bob.Expression] struct {
```
stephenafamo, line 11: The type parameter is unnecessary. Or is it used somewhere?

dialect/mysql/starters.go
```diff
@@ -100,3 +103,30 @@ func Raw(query string, args ...any) Expression {
...
+	cExpr.Whens = append(cExpr.Whens, expr.When{Condition: condition, Then: then})
+	return CaseChain[T](func() expr.Case[T] { return cExpr })
+}
+
+func (c CaseChain[T]) Else(then T) Expression {
+	cExpr := c()
+	cExpr.Else = then
+	var e dialect.Expression
+	return e.New(cExpr)
+}
```
stephenafamo, line 128: Also add an `End` method that returns `Expression`, so that a user can go back to chaining other methods without calling `Else()`

dialect/mysql/starters.go
```diff
@@ -100,3 +103,30 @@ func Raw(query string, args ...any) Expression {
 func Cast(exp bob.Expression, typname string) Expression {
 	return bmod.Cast(exp, typname)
 }
+
+type CaseChain[T bob.Expression] func() expr.Case[T]
```
stephenafamo, line 107: If possible, try to move `CaseChain` to `expr`.

This may involve some type parameter trickery to get it to return `mysql.Expression` from `Else()` and `End()`

However, this will prevent duplication in `psql` and `sqlite`

stephenafamo (changes requested):

k4n4ry:
@stephenafamo 
Thank you for your review. I've fixed your feedback. Could you please take another look?

stephenafamo (approved):

stephenafamo:
One final thing.... add the changes to [CHANGELOG.md](CHANGELOG.md), under the **Unreleased** section.

Also remember to add attribution `thanks (@k4n4ry )`

Thanks for the PR

PR #298, opened 2024-11-10
Add named arguments along with a function to bind to the named args

stephenafamo:


stephenafamo:
@RangelReale  

This is very similar to #104, but with a few differences:

1. `bob.BindNamed` now produces the final query. Merging two steps into one. 
2. The args do not have to be a struct. The type can either be:
    - A struct with fields that match the named arguments in the query
    - A map with string keys. When supplied, the values in the map will be used to bind the named arguments in the query.
    - When there is only a single named argument, one of the following can be used:
        - A primitive type (int, bool, string, etc)
        - `time.Time`
        - Any type that implements `driver.Valuer`.

```go
package main

import (
	"context"
	"fmt"

	"github.com/stephenafamo/bob"
	"github.com/stephenafamo/bob/dialect/psql"
	"github.com/stephenafamo/bob/dialect/psql/sm"
)

type Args struct {
	In1 int
	In2 int
	In3 int
	Id1 int
}

func main() {
	ctx := context.Background()

	query := psql.Select(
		sm.Columns("id", "name"),
		sm.From("users"),
		sm.Where(psql.Quote("id").In(bob.Named("in1", "in2", "in3"))),
	)

	// Use as a regular query
	{
		bound := bob.BindNamed(ctx, query, Args{
			In1: 1,
			In2: 2,
			In3: 3,
			Id1: 4,
		})

		// bound is a [Query] and can be used like any other query

		queryStr, args, err := bob.Build(ctx, bound)
		if err != nil {
			panic(err)
		}

		fmt.Println(queryStr)
		fmt.Println(args)
	}

	// Use in prepared statements
	{
		ctx := context.Background()
		db, err := bob.Open("pgx", "postgresql://user:pass@localhost/bob_droppable?sslmode=disable")

		stmt, err := bob.Prepare[Args](ctx, db, query)
		if err != nil {
			panic(err)
		}

		_ = stmt
	}
}
```

PR #314, opened 2024-11-29
[Proposal] Add transaction management helper

dmakushin:
Suggestion is to add helper method similar to the one described at 
https://bun.uptrace.dev/guide/transactions.html#runintx


stephenafamo:
My first instinct is that this is not useful to Bob. While someone may decide to use a transaction wrapper, there is no need to bake this into Bob.

The reason is that every Bob method take a `bob.Executor`, which is similar to **Bun's IDB interface**. 
1. If you have a top level `bob.DB`, then you can start a transaction with `tx, err := db.BeginTx(ctx, nil)`.
2. If you have a top-level `*sql.DB`:
    ```go
    var db *sql.DB
    tx, err := db.BeginTx(ctx, nil) // tx is *sql.Tx
    bobTxExec := bob.NewTx(tx)
    ```

You can then use this `tx` as your Bob executor, and then commit/rollback as you please (my preferred way is to create a new context and cancel with defer)

In your changes, I can rewrite your `testOperationsWithDatabase` to like this:

```go
func testOperationsWithDatabase(ctx context.Context, db *sql.DB, value int32) error {
	ctx, cancel := context.WithCancel(ctx)
	defer cancel() // rollback transaction if not committed

	sqlTx, err := db.BeginTx(ctx, nil)
	if err != nil {
		return err
	}

	tx := bob.NewTx(sqlTx)

	q := psql.Update(
		um.Table("table"),
		um.Set(psql.Quote("field").EQ(psql.Arg(value))),
	)

	_, err = q.Exec(ctx, tx)
	if err != nil {
		return err
	}

	q2 := psql.Insert(
		im.Into("table2"),
		im.Values(psql.Quote("value").EQ(psql.Arg(value))),
	)

	_, err = q2.Exec(ctx, tx)
	if err != nil {
		return err
	}

	return nil
}
```

dmakushin:
@stephenafamo thank you for the quick response. I understand your opinion. I would like to have such a helper as part of `bob` because it is currently impossible to write a universal helper. The reason for this is that `BeginTx` returns a type specific to bob rather than `sql.Tx`. I understand why this was done, but it forces us to have a helper repository specifically for projects that will be written using `bob`.  As a former user of `bun`, I am used to having such functionality included in the package itself.
Anyway, I will accept an decision. 
Regarding the provided example, I am not sure that I understand how transaction will committed this way. Maybe I am not aware about some auto-commit mechanics which bob has?

stephenafamo:
> Regarding the provided example, I am not sure that I understand how transaction will committed this way. Maybe I am not aware about some auto-commit mechanics which bob has?

My bad 🤦🏾 . I missed calling `tx.Commit()` in my example. Changed the last line to `return tx.Commit()`



> I would like to have such a helper as part of `bob` because it is currently impossible to write a universal helper. The reason for this is that `BeginTx` returns a type specific to bob rather than `sql.Tx`. I understand why this was done, but it forces us to have a helper repository specifically for projects that will be written using `bob`.

🤔 Alright, I see how this could be useful.  
However, instead of placing this in a separate package, how about a method on `bob.DB`.

```go
func (db bob.DB) InTx(ctx context.Context, fn TxFunc, opts ...TxOption) error {
    // ...
}
```

Would this work well for your usecase?

dmakushin:
I see the only benefit of a dedicated package is the ability to register additional callbacks like `OnRollback` and `OnCommit`, as provided in the example. However, in most cases, `db.RunInTx` will be sufficient.
I can prepare changes to add RunInTx as a DB method.

dmakushin:
@stephenafamo I updated PR, but I was unable to test these changes within my code, since the newest version has some breaking changes. After I regenerated the code I see many errors related to missing method `Apply`.
Additionally, conditions like the ones described here - https://bob.stephenafamo.com/docs/models/table#insert - aren't working anymore and now return a query instead of the inserted entity. Were these changes expected?

dmakushin:
Thank you very much for the example. However, this has come as an unpleasant surprise. It will necessitate changes in several services on our side if we want to upgrade the version.

stephenafamo:
I apologise for the breaking changes, it was not made lightly.

One of the main driving factors was making hooks robust. Unfortunately, it was very easy to write/use queries that bypassed hooks and I really want Bob to have very few footguns. If something can be done, it should work as expected.

stephenafamo:
Hello @dmakushin 👋🏾 

Any progress with this?

PR #321, opened 2024-12-05
Fix random value generation for pq.Float64Array factory

felipeparaujo:
The call to [faker's `f.Float64`](https://pkg.go.dev/github.com/jaswdr/faker#Faker.Float64) method was missing parameters. I copied the ones set in this same file for non-array float64s.

To reproduce, create the following table and run `bobgen-psql`:

```
CREATE TABLE test (
  dbl_precision DOUBLE PRECISION [] NOT NULL
);
```

I couldn't find any tests for this behaviour, let me know if there are any that need updating.


stephenafamo:
Thank you. Also update CHANGELOG.md with details of this change under `[Unreleased]`, along with attribution `(thanks @felipeparaujo`

Issue #326, opened 2024-12-13
Missing schema in primary key expression table alias (postgres)


adatob:
Hi,

I noticed an issue using models from Postgres db model genenrator.
During updating or deleting a record using model objects, the WHERE clause based on primary key refers to the non-existing alias. Bellow you can see the generated query:

`UPDATE "schema"."table_name" AS "schema.table_name"
SET "second_column" = $1
WHERE ("table_name"."pk_column" = $2)
RETURNING "schema.table_name"."pk_column" AS "pk_column", "schema.table_name"."second_column" AS "second_column"`

So the problem is probably in generated method `func (o *SchemaTableName) pkEQ() dialect.Expression` returning statement:
`psql.Quote("table_name", "pk_column")`. I'm wondering if there should be also schema name before table name.

best regards,
Adam

PR #327, opened 2024-12-13
Fix missing schema in table alias in pkEQ and pkIN clauses

adatob:
#326


stephenafamo:
Thanks for spotting.

Kindly add a note in `CHANGELOG.md` with attribution `(thanks @adatob)`

PR #328, opened 2024-12-14
Add `net/netip` to random expr imports

plunkettscott:
The `pgtypes.Inet` type uses `net/netip` in its `RandomExpr`, but does not add it to the import list for the `RandomExprImports`. This results in a compilation error for the factory package if no other type exists which also imports the netip package.

I've worked around this in my projects by re-defining the type with the proper import until this can be merged.


stephenafamo:
Kindly add a note in `CHANGELOG.md` with appropriate attribution `(thanks @plunkettscott)`

plunkettscott:
Done

plunkettscott:
Any estimate on when you plan to tag a release with this fix? Having to copy/paste a config across projects right now in the current state to avoid compilation errors.

stephenafamo:
Tagged @plunkettscott

Issue #333, opened 2025-01-02
Allowing to retrieve alias field from the View structure


Nitjsefni7:
In our project, we could benefit from the possibility of retrieving the alias field from the View structure.

The simple method like
`func (v *View[T, Tslice]) Alias() string {
	return v.alias
}
`
would do the trick.

If you approve the proposition, I will be eager to create a pull request.

stephenafamo:
This seems like a straightforward addition.

Can you share what situation you've encountered where you need this?

Nitjsefni7:
When using nested queries in FROM clause, in order to be able to rely on column names provided by bob's models (aliased by default with "schema.table_name") we have to provide the alias to As() method for the nested query. If the Alias() method is provided, we will be able to alias nested queries automatically, based on the table used in the FROM clause of the nested query.

PR #339, opened 2025-01-09
Allowing to retrieve alias field from the View structure

Nitjsefni7:
#333


stephenafamo:
Kindly note down the changes in CHANGELOG under an `UNRELEASED` heading and with attribution (thanks @Nitjsefni7)

PR #371, opened 2025-03-03
Add support for custom types in models directory

jacobmolby:
This adds support for declaring custom types within the "models" directory (or what it is specified as). 

It will make sure the correct imports are used in the corresponding factory files


stephenafamo:
Nice!

This is something I had in mind when I changed the files to use the `.bob.go` prefix and to wipe only bob-specific files.

Fix the formatting, and remember to update `CHANGELOG.md`

Thank you 🙏🏾

jacobmolby:
I have made the changes you suggested now 😄

PR #383, opened 2025-03-29
fix(gen): Support go toolchain directive for version

isgj:
Use the `toolchain` directive from go.mod when detecting the required Go version, falling back to the `go` directive if needed.

Also updates `gofumpt` to version 0.7, which uses
https://pkg.go.dev/go/version#Lang to validate the version.

I was using the new tool option from the go command `go get -tool github.com/stephenafamo/bob/gen/bobgen-psql`  and go resolved one of the dependencies to a newer version `mvdan.cc/gofumpt v0.7.0 // indirect`. This broke the generation of the models, it expects the go version to start with `go` .


stephenafamo:
Kindly update `CHANGELOG.md` with the changes under an `Unreleased` header

PR #389, opened 2025-04-21
Fix not able to retrieve rows

jacobmolby:
Basically we now save the insert expressions, for usage in the `getInserted` method.

Also changed the old tests so they actually mimic the generated code.

About the `if` check in the .tpl, I know it's not ideal, but I didn't really know of any other way, with how it's currently setup.

### DIFF
--- a/dialect/mysql/dialect/insert.go
+++ b/dialect/mysql/dialect/insert.go
@@ -23,10 +23,19 @@ type InsertQuery struct {
 	ColumnAlias        []string
 	Sets               []Set
 	DuplicateKeyUpdate clause.Set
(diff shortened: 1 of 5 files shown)


stephenafamo:
Thanks for all your work 🙏🏾 

I may take some time to review this but feel free to remind me if I haven't done so in a week

stephenafamo:
I apologise for making you do so much work.  
I found some time to investigate the underlying issue and the fix I ended up with was different.  

Try out this commit 4e3e86b, I believe the issue is fixed there.

jacobmolby:
Hey man, no worries. There is no doubt your fix is much better. It seems to work. Thanks!

PR #390, opened 2025-04-24
Fix enum types match by "db_type" when using bobgen-psql

abramlab:
When using enum types, the `Column.DBType` now correctly includes the schema name and replacement works well.

Examples:

// migrations.up.sql
```sql
CREATE SCHEMA IF NOT EXISTS preferences;
CREATE TYPE preferences.test AS ENUM ('foo', 'bar');
CREATE TABLE IF NOT EXISTS preferences.users
(
    id SERIAL PRIMARY KEY,
    name VARCHAR(255) NOT NULL,
    test preferences.test NOT NULL
);
```
// bobgen.yaml
```yaml
...
types:
  mtype.Type:
    no_randomization_test: true
    imports:
      - '"test-bob/mtype"'

replacements:
  - tables: []
    match:
      db_type: "preferences.test"
    replace: "mtype.Type"
```

Note: If the PostgreSQL type is created without an explicit schema (e.g., `CREATE TYPE test AS ENUM ('foo', 'bar');`), it defaults to the public schema. In this case, the replacement config should explicitly specify `public`:
```yaml
match:
  db_type: "public.test"
```

### DIFF
--- a/gen/bobgen-psql/driver/exclude-tables.golden.json
+++ b/gen/bobgen-psql/driver/exclude-tables.golden.json
@@ -486,7 +486,7 @@
 				},
 				{
 					"name": "enum_use",
-					"db_type": "ENUM",
+					"db_type": "public.unicode_enum",
 					"default": "",
 					"comment": "",
 					"nullable": false,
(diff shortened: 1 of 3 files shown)


stephenafamo:
You'd need to update the golden files

```
go test ./gen/bobgen-psql/driver -overwrite-golden
```

Issue #194, opened 2024-03-14
Invalid SQL for In operator with empty slice values.


atzedus:
```go
vals1 := []bob.Expression{}
fmt.Println(psql.Quote("id").In(vals1...).String())
// ("id" IN )

vals2 := []any{}
fmt.Println(psql.Quote("id").In(psql.Arg(vals2...)).String())
// ("id" IN ())
```

Is this expected behavior or is it a bug?

In sqlboiler similar expressions generates code like `("id" IN (1=0))`

stephenafamo:
I think it is fair that trying to compare against an empty slice will be invalid. Maybe the better thing would be to throw an error instead 🤔 .

Can you help me understand why you would expect this to generate `("id" IN (1=0)`, or if you think another result would be more appropriate, let me know.

jacobmolby:
@stephenafamo From SQLBoiler's code:
```
// WHERE IN () is invalid sql, so it is difficult to simply run code like:
// for _, u := range model.Users(qm.WhereIn("id IN ?",uids...)).AllP(db) {
//    ...
// }
// instead when we see empty IN we produce 1=0 so it can still be chained
// with other queries
```

https://github.com/volatiletech/sqlboiler/blob/master/queries/query_builders.go#L394C4-L399C25

stephenafamo:
Wouldn't this be a little dangerous?

* `1=0` does not really mean the SQL is valid. It would fail if it is not being compared to a boolean. (at least in Postgres)
* If it is being compared to a boolean, this could lead to subtle bugs because this really means `WHERE id IN (false)` which may not be what the user expects.

I am not sure this is the better user experience.

atzedus:
So, i try to explain source of my question.

From my experience, in many other ORM this is expected behavior.

For example (Eloquent ORM - Laravel (php)):
```php
DB::table('products')->whereIn('id', [])->get()
// SQL: select * from `products` where 0 = 1
```

Bun (golang)
```go
q = q.Where("user_id IN (?)", bun.In([]int64{}))
// SQL: user_id IN (NULL)
```

GORM (golang)
```go
db.Where("user_id IN ?", []string{}).Find(&products)
// SQL: SELECT * FROM `products` WHERE user_id IN (NULL)
```

So, sqlboiler! :) Currently we in migration process from sqlboiler to bob, and this is why I did not expect this behavior.

stephenafamo:
[...]

PR #392, opened 2025-04-24
Ability to use 'In' and 'NotIn' where mods with empty value slices

abramlab:
Added the ability to use `In` and `NotIn` where modifiers with empty value slices.

- Follows the same pattern as [sqlboiler's handling of empty IN clauses](https://github.com/volatiletech/sqlboiler/commit/569dd00c27fa61c96c708db179f87bab794d8b7c#diff-1903c797de1900cb84322838dabf94209eea5277d3b7f12c03d9895d6d8b4484R395-R399)
- Empty `In` clauses will resolve to FALSE (no results)
- Empty `NotIn` clauses will resolve to TRUE (all results)

Thanks to @Rubanik-Alexei for suggesting this approach.

### DIFF
--- a/dialect/mysql/where.go
+++ b/dialect/mysql/where.go
@@ -2,6 +2,7 @@ package mysql
 
 import (
 	"github.com/stephenafamo/bob"
+	"github.com/stephenafamo/bob/expr"
 	"github.com/stephenafamo/bob/mods"
 )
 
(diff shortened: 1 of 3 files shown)


stephenafamo:
Thank you for sending in this PR. However, I will not be merging it because:

1. I think using `1=0` or `1=1` can lead to subtle bugs if evaluated against booleans [see comment](https://github.com/stephenafamo/bob/issues/194#issuecomment-1997661501).
2. Instead, similar to #195, I've made empty args evaluate to `NULL` in 255a1d3

This means, both `In(nil)` and `NotIn(nil)` will both evaluate to `NULL` and no rows will be returned.  
I think this is more "correct"


Thank you once again for the PR and bringing this issue to my attention.

PR #431, opened 2025-05-28
Add ncruces/go-sqlite3 support

ncruces:
As discussed on Reddit, and as a first approximation, this is what should be required to support my driver.

What's the simplest, but still realistic, way to test this? I ask because my driver has a few peculiarities, and I'd like to be sure I'm not messing anything up.

The one that matters the most here, is that the driver is split into 3 packages:
- [`github.com/ncruces/go-sqlite3`](https://pkg.go.dev/github.com/ncruces/go-sqlite3) wraps the [C SQLite API](https://sqlite.org/cintro.html) - this is needed for the `sqlite3.Error` type
- [`github.com/ncruces/go-sqlite3/driver`](https://pkg.go.dev/github.com/ncruces/go-sqlite3/driver) registers the `database/sql` driver
- [`github.com/ncruces/go-sqlite3/embed`](https://pkg.go.dev/github.com/ncruces/go-sqlite3/embed) embeds/links SQLite into your application

You don't need to import `driver` or `embed` everywhere in your project, and in fact I'd argue if you're creating (or in this case generating) a library you _shouldn't_ import them at all.

The reason is that, e.g., for the `embed` package in particular, you may decide to embed/link a modified SQLite build. I have an [alternative package](https://github.com/ncruces/go-sqlite3/tree/main/embed/bcw2#embeddable-wasm-build-of-sqlite) that includes "experimental" SQLite features, Alex [publishes](https://alexgarcia.xyz/sqlite-vec/go.html#ncruces) another one that supports vector search, etc.

It's up to the final end user which build of SQLite they want to link with, and they can decide by importing the correct package in the `main.go` of their application (or some other place where it makes sense).

I'm sorry if this over complicates things.

### DIFF
--- a/gen/bobgen-sqlite/driver/sqlite_test.go
+++ b/gen/bobgen-sqlite/driver/sqlite_test.go
@@ -136,7 +136,7 @@ func TestAssembleSQLite(t *testing.T) {
 
 	config := Config{
 		Config: helpers.Config{
-			Dsn: mainDB.Name() + "?_pragma=foreign_keys(1)&_pragma=busy_timeout(10000)",
+			Dsn: "file:" + mainDB.Name() + "?_pragma=foreign_keys(1)&_pragma=busy_timeout(10000)",
 		},
 		Attach: map[string]string{"one": oneDB.Name()},
 	}
@@ -172,6 +172,10 @@ func testSQLiteDriver(t *testing.T, config Config) {
 		{
 			name:       "modernc",
 			driverName: "modernc.org/sqlite",
+		},
+		{
+			name:       "ncruces",
+			driverName: "github.com/ncruces/go-sqlite3",
 		},
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added the `drivers/pgx` package which contains types and functions to use a `pgx/v5` connection as a `bob.Executor`.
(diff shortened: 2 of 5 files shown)


stephenafamo:
Right now, the way to test would be to setup a project locally, generate code for an example schema (that includes unique constraints) and then run the generated tests.

You will have to supply a `testDB` so that the tests requiring a database connection are not skipped.  
You can use `test/gen/templates/database.bob_test.go.tpl` as an example, but this template is only added during tests.

I should extend the tests to run for each driver combination, right now it only runs for the default drivers.

Speaking of, I'm not sure if your driver supports attaching databases and how that should ideally be done.

gen/bobgen-sqlite/driver/sqlite_test.go
```diff
@@ -136,7 +136,7 @@ func TestAssembleSQLite(t *testing.T) {
 
 	config := Config{
 		Config: helpers.Config{
-			Dsn: mainDB.Name() + "?_pragma=foreign_keys(1)&_pragma=busy_timeout(10000)",
+			Dsn: "file:" + mainDB.Name() + "?_pragma=foreign_keys(1)&_pragma=busy_timeout(10000)",
```
ncruces, line 139: My driver is perhaps too strict, but it follows [SQLite in requiring the `file:` scheme](https://sqlite.org/uri.html#uri_format):
> The scheme is always required. (...) The scheme of the URI must be "file:".

Also [here](https://www.sqlite.org/c3ref/open.html#:~:text=URI%20Filenames,-If%20URI%20filename):
> If [URI filename](https://www.sqlite.org/uri.html) interpretation is enabled, and the filename argument begins with "file:" (...)

PR #459, opened 2025-06-12
implement new TypeModifier to use aarondl/opt/null package for nullab…

Nitjsefni7:
…le values and pointers for optional values, fixes to pointers optional type method

### DIFF
--- a/gen/drivers/types.go
+++ b/gen/drivers/types.go
@@ -371,6 +371,29 @@ func (a AarondlNull) NullTypeImports(forType Type) []string {
 	return append(slices.Clone(forType.Imports), `"github.com/aarondl/opt/null"`)
 }
 
+var _ TypeModifier = AarondlNullPointers{}
+
+type AarondlNullPointers struct{}
+
+func (a AarondlNullPointers) NullType(name string) NullType {
+	return NullType{
+		Name:              fmt.Sprintf("null.Val[%s]", name),
+		ValidExpr:         "SRC.IsValue()",
+		UseExpr:           "SRC.MustGet()",
+		UseExprImports:    []string{},
+		CreateExpr:        "null.From(SRC)",
+		CreateExprImports: []string{`"github.com/aarondl/opt/null"`},
+	}
+}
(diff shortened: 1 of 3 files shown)


gen/drivers/types.go
```diff
@@ -447,11 +479,10 @@ func (d DatabaseSqlNull) OptionalType(name string, def Type, isNull, fromOrToNul
 			ot.CreateExprImports, def.NullType.UseExprImports...,
 		)
 
+	default:
+		ot.UseExpr = "*SRC"
+		ot.CreateExpr = "&SRC"
```
stephenafamo, line 484: This can potentially cause issues in certain cases. Best to leave it as the closures already defined at the top of the function.

stephenafamo: > but I had suspicion, that it is a bit too wide of a case to handle it using just the default option.

This is correct. The main reason for using the closure for pointers is for cases where we cannot directly take the address or dereference `SRC`. For example, if the create/use expression of a null type uses a type conversion, it will create invalid code:

The type conversion in the following code is unnecessary, but it illustrates a possibility

```go
// Invalid code
	video1.SponsorID = &sql.NullInt32(sql.NullInt32{Int32: int32(sponsor0.ID), Valid: true})

// Valid code
	video1.SponsorID = func() *sql.NullInt32 {
		v := sql.NullInt32(sql.NullInt32{Int32: int32(sponsor0.ID), Valid: true})
		return &v
	}()
```

While this creates somewhat unnecessarily verbose code, I believe the compiler is smart enough to inline it, so there should be no performance penalties.

> Have you considered creating test cases for the examples that you are aware are causing issues to verify if optional, null etc. create and use expression are generated correctly?

Test cases will be very welcome.

Perhaps we can use [go/parser.ParseExpr](https://pkg.go.dev/go/parser#ParseExpr) to verify that the syntax is valid.

stephenafamo:
Kindly resolve the conflicts when you can @Nitjsefni7 🙏🏾

PR #469, opened 2025-07-01
chore: update docs for readability, grammar, and consistency

singhsays:
Minor language, grammar and consistency fixes for documentation. Also, included code generation runner details on the intro page as thats where the reader would expect to find them initially.

### DIFF
--- a/website/docs/README.md
+++ b/website/docs/README.md
@@ -28,15 +28,14 @@ Bob's philosophy centers around the following:
 
 ## 1. Query Builder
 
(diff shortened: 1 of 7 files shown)


singhsays:
This is likely related to golangci-lint updating the noctx linter and since its not pinned to a specific version, this probably got updated recently to latest v2.2.0 that includes related changes https://github.com/golangci/golangci-lint/blob/main/CHANGELOG.md#v220

singhsays:
Yup, seems like https://github.com/sonatard/noctx/pull/37/files which was merged last week.

We could either disable the linter for now or update to make the related methods ctx aware. WDYT?

singhsays:
Since we were already passed a context, it was simpler to just pass the context through. Added a fixed, this should unbreak CI. Updated the changelog as well.

PR #470, opened 2025-07-01
Add FromExisting<Rel> and WithExisting<Rel> to factory

dutow:
### Use case

I'm trying to setup a more complex sample/test database using factories. I already have persisted records for some of the schemas (using earlier calls to factories), and now I'm trying to create new records that refer to the existing entries instead of creating new records.

For example think about a forum: it has users and threads. There are already users in the database, and now we want to create threads with the creator set to a random existing user.

Something like this works if the relation is not required:

```
findRandomUser := func() int32 {
		res, _ := models.Users.Query(sm.Columns("id"), sm.OrderBy("random()")).One(ctx, bobdb)
		return res.ID
	}
	threadTemplate := factoryCtx.NewThread(ctx, factory.ThreadMods.CreatorUserIDFunc(findRandomUser))

```

But it doesn't work with required relations, because in that case the factory always creates a new relation.

This PR basically unifies the behavior for the two cases.

### Alternative idea

I was also thinking about proposing a new method for relations, `WithExisting<RelationName>`. But the problem with this is:
* it needs a way to properly select which record(s) to use (always using random doesn't seem right)
* it would leave the inconsistency I mentioned above

So instead I'm proposing this patch to unify the behavior.

What do you think?

(`m.R.relation` won't be set in this case - but that's already the case for optional relations when done this way)

### DIFF
--- a/gen/templates/factory/bobfactory_main.bob.go.tpl
+++ b/gen/templates/factory/bobfactory_main.bob.go.tpl
@@ -1,4 +1,5 @@
 {{$.Importer.Import "context"}}
+{{$.Importer.Import "models" (index $.OutputPackages "models") }}
 
 type Factory struct {
     {{range $table := .Tables}}
@@ -23,6 +24,19 @@ func (f *Factory) New{{$tAlias.UpSingular}}(ctx context.Context, mods ...{{$tAli
   {{$tAlias.UpSingular}}ModSlice(mods).Apply(ctx, o)
 
 	return o
+}
+
+func (f *Factory) FromExisting{{$tAlias.UpSingular}}(m models.{{$tAlias.UpSingular}}) *{{$tAlias.UpSingular}}Template {
+	o := &{{$tAlias.UpSingular}}Template{}
+
+  {{range $column := $table.Columns -}}
+  {{$colAlias := $tAlias.Column $column.Name -}}
+  {{- $colTyp := $.Types.GetNullable $.CurrentPackage $.Importer $column.Type $column.Nullable -}}
+        o.{{$colAlias}} = func() {{$colTyp}} { return m.{{$colAlias}} }
+  {{end}}
+  o.alreadyPersisted = true
+
+  return o
 }
 
 {{end}}
--- a/gen/templates/factory/table/01_types.go.tpl
+++ b/gen/templates/factory/table/01_types.go.tpl
@@ -36,6 +36,8 @@ type {{$tAlias.UpSingular}}Template struct {
         r {{$tAlias.DownSingular}}R
     {{- end}}
     f *Factory
+
+    alreadyPersisted bool
 }
 
 {{if $.Relationships.Get $table.Key -}}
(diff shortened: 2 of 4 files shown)


stephenafamo:
On first look, this looks good.

However, I'm more in favour of your **Alternative Idea**.   
I don't understand why a potential `WithExisting<RelationName>` will need to select which record to use and what inconsistency it will cause.

I assume that it will take the relation that should be used.. for example:


```go
findRandomUser := func() *models.User {
		res, _ := models.Users.Query(sm.Columns("id"), sm.OrderBy("random()")).One(ctx, bobdb)
		return res.ID
	}
	threadTemplate := factoryCtx.NewThread(ctx, factory.ThreadMods.WithExistingUser(findRandomUser()))
```

### Yet another suggestion

In your test, why not create the user, with all the threads

```go
user, _ := factoryCtx.NewUser(ctx, factory.UserMods.WithNewThreads(5))

threads := user.R.Threads
// Use threads here
```

stephenafamo:
This is really nice. Great work 🙏🏾 

The main suggestion I have will be to also account for relationships that exist in the model passed to `FromExisting<Relation>` to also create relations in the template using `WithExisting`.

I think the earlier changes that are based on the column name can be dropped since this does the same thing... but better 😁

gen/templates/factory/table/02_build.go.tpl
```diff
@@ -101,6 +101,17 @@ func (o {{$tAlias.UpSingular}}Template) Build() *models.{{$tAlias.UpSingular}} {
 	return m
 }
 
+func (o *{{$tAlias.UpSingular}}Template) From(locator func() models.{{$tAlias.UpSingular}}) {
```
stephenafamo, line 104: 1. Why is `locator` a function instead of the model?
2. I think this method should not be on the template, but should be a method of the factory directly and will construct and return a new template similar to `f.New<ModelName>`.   
This makes it unlikely that the templated values will be different from the model itself

stephenafamo (changes requested):

gen/templates/factory/table/02_build.go.tpl
dutow: 1: for example, in the example I used (findRandomUser), if we create multiple models with `CreateMany`, wouldn't it make sense to use different random users? This is why I went with the function, but for any other use case I think of, a model would be fine. I can change it to that if you think it is better.

(there's a separate issue currently that CreateMany only creates relations for the first returned model, it's empty for all others - but assuming that's fixed)

dutow: Updated: method removed, and the other methods now accept a model directly.

stephenafamo (approved):

stephenafamo:
You may want to take a look at 1e767eb. It does the following

* Adds a `AddExisting<Rel>` for to-many relationships
* Adds any loaded relationships into the template in `FromExisting<rel>`
* Account for `alreadyPersisted` in `insertOptRels`

PR #495, opened 2025-07-16
Update changelog with previous fix and fix mysql tests

luiscleto:
Previous contribution broke tests with include/exclude tables.
This PR adds a `queries` params to tests and includes it only in the base test with all tables.
This also adds another test with a type annotation in a query to be updated if/when behavior changes in case this is not the intended behavior

Also updates changelog with previous fix for backticks around table names in mysql queries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix transformer for single result queries.
 - Properly handle indexes where the sorting order can be null.
 - Check for nullability when loading relationships.
-- Handle table names quoted with backticks in mysql query parser.
+- Handle table names quoted with backticks in mysql query parser. (thanks @luiscleto)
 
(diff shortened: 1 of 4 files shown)


stephenafamo:
> There seems to be some (unrelated?) flakyness with the multi_keys table

Yeah... I need to find some time to dig into it 😫

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix transformer for single result queries.
 - Properly handle indexes where the sorting order can be null.
 - Check for nullability when loading relationships.
+- Handle table names quoted with backticks in mysql query parser.
```
stephenafamo, line 43:
```suggestion
- Handle table names quoted with backticks in mysql query parser. (thanks @luiscleto)
```

stephenafamo (changes requested):

stephenafamo (approved):

Issue #497, opened 2025-07-17
Allow matching columns by regex in code generation config


abdusco:
Currently, the replacements configuration only supports exact matching for column properties such as `name`, `db_type`, etc. This limits flexibility, for example, when attempting to replace all foreign key ID columns `(*_id)` across multiple tables (e.g. `authors.id` <- `books.author_id`).

```yaml
replacements:
  - tables: ["table_name"] # What tables to look inside. Matches all tables if empty
    match:
      name: "username" # Matches the column name
      db_type: "varchar(255)" # Matches the database type
      default: "NULL" # Matches the default value
      comment: "The username" # Matches the column comment
      nullable: true # Matches the nullable value. Defaults to false.
      generated: false # Matches the generated value. Defaults to false.
      autoincr: false # Matches the autoincr value. Defaults to false.
```

I'd like to request adding regex support for the matching fields (e.g. `name`) in the replacements configuration, similar to how regex is supported in the `only` and `except` options described here: https://bob.stephenafamo.com/docs/code-generation/psql#onlyexcept

For example, this would allow:
```yaml
replacements:
  - match:
      name: "/id$/"
      db_type: integer
    replace: "ID"
```

I’m happy to help implement this feature if desired!

PR #499, opened 2025-07-18
feat(gen): Support matching columns case-insensitively or by regex in replacements

abdusco:
Closes: #497

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,12 +20,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add WithExisting<Rel> to factory mods to attach an existing model as a relationship. (thanks @dutow)
 - Added support in psql for combined args (order by, limit etc.) in combined queries and use parens if they are present. (@iwyrkore)
 - Added parens for combined queries. (@iwyrkore)
+- Match columns using regular expressions in type replacements. This is useful for e.g. matching columns that have a common prefix or suffix. (thanks @abdusco)
 
 ### Changed
 
 - `Mod` is now a separate field in `orm.ModQuery` and `orm.ModExecQuery`.
 - `Allx` now takes a `Transformer` type parameter to transform the result of the query.
 - Updated documentation for readability, added code gen examples. (thanks @singhsays)
+- Columns are now matched in a case-insensitive manner in type replacements. (thanks @abdusco)
 
 ### Removed
 
@@ -41,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Properly handle indexes where the sorting order can be null.
 - Check for nullability when loading relationships.
 - Handle table names quoted with backticks in mysql query parser. (thanks @luiscleto)
+- Allow matching columns in type replacements by the `autoincr` property as stated in the docs. (thanks @abdusco)
 
 ## [v0.38.0] - 2025-06-04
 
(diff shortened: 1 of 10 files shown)


gen/types.go
```diff
@@ -101,6 +129,10 @@ func matchColumn(c, m drivers.Column) bool {
 		return false
 	}
 
+	if m.AutoIncr != c.AutoIncr {
+		return false
+	}
```
abdusco, line 134: This wasn't being checked before despite the docs claimed that it was.

```md
## Replacements

There exists the ability to override types that the driver has inferred. The way to accomplish this is through the config file.

```yaml
replacements:
  - match:
      ...
      autoincr: false # Matches the autoincr value. Defaults to false.
```

gen/types.go
```diff
@@ -55,6 +57,28 @@ func processTypeReplacements[C, I any](types drivers.Types, replacements []Repla
 	}
 }
 
+// matchString reports whether string a matches a pattern.
+// Pattern a can be either a literal string (case-insensitive comparison)
+// or a regular expression enclosed with /. Regex patterns are automatically made case-insensitive.
+func matchString(pattern, candidate string) bool {
```
abdusco, line 63: Everything is matched case-insensitively now. Is that alright?

gen/types.go
```diff
@@ -55,6 +57,28 @@ func processTypeReplacements[C, I any](types drivers.Types, replacements []Repla
...
+	stringPatterns, regexPatterns := drivers.ClassifyPatterns([]string{pattern})
+	for _, pattern := range stringPatterns {
+		if strings.ToLower(pattern) == strings.ToLower(candidate) {
```
stephenafamo, line 66:
```suggestion
		if strings.EqualFold(pattern, candidate) {
```

stephenafamo (changes requested):
Great work 👍🏾 

Also add a note in the CHANGELOG.md along with attribution (thanks @abdusco)

gen/types.go
abdusco: I thought about this but this also allows `ß` and `ss` to match, so I'm not sure.

gen/types.go
stephenafamo: Also in practice, I don't think it is reasonable to have a "Users" and "users" table in the same database

gen/types.go
abdusco: No, I actually misread the docs

https://pkg.go.dev/strings#EqualFold

> `fmt.Println(strings.EqualFold("ß", "ss"))  // false because comparison does not use full case-folding`

abdusco:
Addresses the requests.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add WithExisting<Rel> to factory mods to attach an existing model as a relationship. (thanks @dutow)
 - Added support in psql for combined args (order by, limit etc.) in combined queries and use parens if they are present. (@iwyrkore)
 - Added parens for combined queries. (@iwyrkore)
+- Match columns in a case-insensitive manner and also using regular expressions in replacements. This is useful for e.g. matching columns that have a common prefix or suffix. (@abdusco)
```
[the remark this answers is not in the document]
stephenafamo: Yes, under each header as appropriate

abdusco:
Adjusted the changed log and fixed the liner error 👍.

stephenafamo (approved):

Issue #500, opened 2025-07-19
Replacement config should ignore unset fields


abdusco:
This is a follow-up to https://github.com/stephenafamo/bob/issues/497.

It's worth considering not using the same `drivers.Column` type for filtering columns and modelling database columns in `gen.matchColumn`:

https://github.com/abdusco/bob/blob/2fbab61e2ec6e14c0f188d66a9452a975e0ddc91/gen/types.go#L126-L132

If a bool field is not specified in the config, it should not be considered as explicitly set to `false`. Currently, `autoincr: false` is treated the same as if it were omitted, due to Go's zero value semantics, which is a surprising behavior. I wouldn't expect a field I don't even have in my config file to be used for evaluation.

I propose having another type for this with pointer fields, so the omitted fields in the config can be handled properly:

```go
type ColumnFilter struct {
  Name *string
  DBType *string
  ...
  Generated *bool 
  AutoIncr *bool
}
```

(I'd be happy to implement this)

stephenafamo:
This should only be for boolean fields right? Since we can check if the string is empty or not for string fields.

Also, we can use `github.com/aarondl/opt/omit` instead of pointers

abdusco:
> github.com/aarondl/opt/omit

Very good point, it's a dependency we're already fetching 👍.

> only for boolean fields

Correct. But for consistency, we could mark all fields as `omit.Value`

abdusco:
NOTE: This also means that the current behavior of matching bool fields only when a string field is specified, and the side effect where bool columns are always evaluated, is redundant. This way, one can match all `autoincr` fields if they want, or match any field without duplicating rules for columns with and without autoincr.

abdusco:
Something about the test `bobgen-mysql.TestDriver.default` is flaky.

stephenafamo:
Yeah, it's been flaky for a while, I need to find time to fix it

PR #501, opened 2025-07-20
fix(gen): Replacement configuration should only consider fields that are set

abdusco:
Closes: #500


stephenafamo:
Great work once again 🙏🏾

Issue #502, opened 2025-07-20
Dashes in enums are translated as u2d by the code generator


abdusco:
Dashes in enums are converted to Unicode. They should be treated the same as underscores.

```diff
const (
-	PaymentTypeOneU2doff   PaymentType = "one-off"
-	PaymentTypeOneOff   PaymentType = "one-off"
)
```

This should be an easy fix (I'll submit a PR)

https://github.com/stephenafamo/bob/blob/f2f2af82925cb032944eda56030a91173cb0af93/gen/templates.go#L111

PR #503, opened 2025-07-20
fix(gen): Improve enum value generation

abdusco:
- Treat dashes in enum values the same as underscores
- Ignore spaces
- Normalize uppercase values to TitleCase

Closes: #502


abdusco:
This is potentially a breaking change. But I guess that's OK, because the package is at v0 🤷

stephenafamo:
Thanks again 👍🏾

Issue #504, opened 2025-07-21
Improve factory API


abdusco:
Right now, to create a new record, one has to pass the ctx twice, e.g. `f.NewX(ctx, ...).MustCreate(ctx, exec)`. 

Could this be improved by propagating `ctx` passed to the `MustCreate` method to the entity template, so we could have `f.NewX(...).MustCreate(ctx, exec)`?

stephenafamo:
This makes sense.

abdusco:
Should I go ahead and work on this? Do you have any tips before I dive in?

stephenafamo:
I've been reviewing the generated code, and I don't think this is possible. The contexts are used for different things.

The first context is passed to the mods. The second context is used for the database query.

There is no good way to eliminate either. I'm open to suggestions

abdusco:
We could update update `New` to use an internal `context.Background()` and introduce a new `NewWithContext` method that accepts a ctx from the caller

stephenafamo:
That works 👍🏾 

Happy to review a PR

PR #506, opened 2025-07-21
feat(gen): Support creating templates without having to pass a context

abdusco:
Closes: #504

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated documentation for readability, added code gen examples. (thanks @singhsays)
 - Columns are now matched in a case-insensitive manner in type replacements. (thanks @abdusco)
 - Columns can now be matched with as many conditions as needed in type replacements. This removes the previous requirement that boolean fields had to be specified in addition to a string field. (thanks @abdusco)
+- Factories now expose a `NewXWithContext` that accepts a context in addition to `NewX` that does not. This provides a cleaner API for the callers, while still allowing the use of context internally. (thanks @abdusco)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Check for nullability when loading relationships.
 - Handle table names quoted with backticks in mysql query parser. (thanks @luiscleto)
 - Allow matching columns in type replacements by the `autoincr` property as stated in the docs. (thanks @abdusco)
+- Handle dashes and spaces in generated enum values properly (thanks @abdusco)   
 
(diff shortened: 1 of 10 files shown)


abdusco:
Please don't merge right away, I'll update README + CHANGELOG after you OK this.

abdusco:
Ah yes, I had `no_tests: true` in my config.

gen/templates/factory/bobfactory_main.bob_test.go.tpl
```diff
@@ -22,7 +22,7 @@ func TestCreate{{$tAlias.UpSingular}}(t *testing.T) {
     }
   }()
 
-  if _, err := New().New{{$tAlias.UpSingular}}(ctx).Create(ctx, tx); err != nil {
+  if _, err := New().New{{$tAlias.UpSingular}}WithContext(ctx).Create(ctx, tx); err != nil {
```
abdusco, line 25: Should the tests call `.New` or `.NewWithContext`? I think the former. Even though a ctx is available, passing it to the builder feels redundant.

stephenafamo: If a context is available, let's use it.

Actually, since Bob targets Go 1.24 at minimum, we should use [`t.Context()`](https://pkg.go.dev/testing#T.Context) instead of `context.Background()`

abdusco: Alright, I'll change `context.Background()` to `t.Context()` too.

abdusco:
I had forgotten to add a test case unicode enum values in #502 

I squeezed in another test here. I hope you don't mind it.

https://github.com/stephenafamo/bob/pull/506/commits/5dec281e19106d2f7dc516b528767a08157cee02

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated documentation for readability, added code gen examples. (thanks @singhsays)
 - Columns are now matched in a case-insensitive manner in type replacements. (thanks @abdusco)
 - Columns can now be matched with as many conditions as needed in type replacements. This removes the previous requirement that boolean fields had to be specified in addition to a string field. (thanks @abdusco)
+- Factories now expose a `NewXWithContext` that accepts a context in addition to `NewX` that does not. This provides a cleaner API for the callers, while still allowing the use of context internally. (thanks @abdusco)
```
abdusco, line 32: Turns out the docs never mentioned the use of ctx

https://github.com/stephenafamo/bob/blob/2f1f83bba3c103d5988f3b4774384aaf9986e86d/website/docs/code-generation/factories.md?plain=1#L24

abdusco:
Turns out the docs never mentioned the use of ctx, so I won't touch it for now.

https://github.com/stephenafamo/bob/blob/2f1f83bba3c103d5988f3b4774384aaf9986e86d/website/docs/code-generation/factories.md?plain=1#L24

PR #511, opened 2025-07-23
fix(gen): check for nullability when loading relationships from Slices

### LINKED ISSUE #492: SelectThenLoad panics for nullable nested relationships
This issue is similar to https://github.com/stephenafamo/bob/issues/477 but not the same.

With following schema and entries

```
CREATE TABLE public."content" (
	id uuid NOT NULL,
	source_id uuid NULL,
	CONSTRAINT content_pkey PRIMARY KEY (id)
);

CREATE TABLE public.sources (
	id uuid NOT NULL,
	CONSTRAINT sources_pkey PRIMARY KEY (id)
);

CREATE TABLE public.source_files (
	id uuid NOT NULL,
	source_id uuid NOT NULL,
	CONSTRAINT source_files_pkey PRIMARY KEY (id)
);


INSERT INTO public."content" (id,source_id) VALUES
         ('722253fd-fea2-4435-98a8-e57dbc051793'::uuid,NULL),
         ('33137c38-9bc6-44d0-87a0-3f66a481d6a0'::uuid,'d474da7f-50da-44f5-9594-91f5818d1312'::uuid);

INSERT INTO public.sources (id) VALUES
         ('d474da7f-50da-44f5-9594-91f5818d1312'::uuid);

INSERT INTO public.source_files (id,source_id) VALUES
         ('30b80d3a-44ca-428f-8d4d-63f3cb28c130'::uuid,'d474da7f-50da-44f5-9594-91f5818d1312'::uuid);
```

Running the following script after generating the code using bobgen-psql built from commit cff2063d33648e9a79f6336478f393705c1f8705

```
package main

import (
	"context"
	"database/sql"
	"fmt"

	"github.com/stephenafamo/bob"
)

func main() {
	conn, err := sql.Open("pgx", "postgres://pg:dummy@localhost:13524/bug_repro?sslmode=disable")
	if err != nil {
		panic(err)
	}

	db := bob.NewDB(conn)

	contents, err := Contents.Query(
		Preload.Content.Source(
			SelectThenLoad.Source.SourceFiles(),
		),
	).All(context.Background(), db)
	if err != nil {
		panic(err)
	}

	fmt.Println(contents)
}
```

panics with

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x104959cac]

main.SourceSlice.SourceFiles({0x140000c8ad0, 0x2, 0x14000035a78?}, {0x0, 0x0, 0x0})
        repro/sources.bob.go:455 +0x5c
main.SourceSlice.LoadSourceFiles({0x140000c8ad0, 0x2, 0x104ae1940?}, {0x104b8d0b0, 0x104f19840}, {0x104b88fa8, 0x140000bcc30}, {0x0?, 0x14000035b88?, 0x104684508?})
        repro/sources.bob.go:624 +0x60
main.buildSourceThenLoader[...].func2({0x104b88fa8?, 0x140000bcc30?}, {0x1051b1c30?, 0x140000f24e0?}, {0x0?, 0x140000c8ad0?, 0x16?})
        repro/sources.bob.go:534 +0x5c
main.buildSourceThenLoader[...].func4.1({0x104b88fa8?, 0x140000bcc30?}, {0x104b5d8c0?, 0x140000f24e0?})
        repro/bob_main.bob.go:116 +0xdc
github.com/stephenafamo/bob/orm.Loader[...].Load(...)
        /go/pkg/mod/github.com/stephenafamo/bob@v0.38.1-0.20250714151549-cff2063d3364/orm/load.go:28
github.com/stephenafamo/bob/orm.(*AfterPreloader).Load(0x140000f8960?, {0x104b8d0b0, 0x104f19840}, {0x104b88fa8, 0x140000bcc30}, {0x14000035e40?, 0x14000230000?})
        /go/pkg/mod/github.com/stephenafamo/bob@v0.38.1-0.20250714151549-cff2063d3364/orm/load.go:178 +0x114
github.com/stephenafamo/bob.Allx[...]({0x104b8d0b0, 0x104f19840}, {0x104b88fa8, 0x140000bcc30}, {0x104b8bbd8, 0x140000a3b60}, 0x140000a3800)
        /go/pkg/mod/github.com/stephenafamo/bob@v0.38.1-0.20250714151549-cff2063d3364/exec.go:180 +0x36c
github.com/stephenafamo/bob/orm.Query[...].All(...)
        /go/pkg/mod/github.com/stephenafamo/bob@v0.38.1-0.20250714151549-cff2063d3364/orm/query.go:68
main.main()
        repro/main.go:24 +0x150
exit status 2
```

felipeparaujo:
Fixes #492

### DIFF
--- a/gen/templates/models/table/10_rel_load.go.tpl
+++ b/gen/templates/models/table/10_rel_load.go.tpl
@@ -239,11 +239,19 @@ func (os {{$tAlias.UpSingular}}Slice) Load{{$relAlias}}(ctx context.Context, exe
 
 	{{if $rel.IsToMany -}}
 		for _, o := range os {
+			if o == nil {
+				continue
+			}
+
 			o.R.{{$relAlias}} = nil
 		}
 	{{- end}}
 
 	for _, o := range os {
+		if o == nil {
+			continue
+		}
+
 		for _, rel := range {{$fAlias.DownPlural}} {
 			{{range $index, $local := $side.FromColumns -}}
         {{- $foreign := index $side.ToColumns $index -}}
@@ -312,7 +320,7 @@ func (os {{$tAlias.UpSingular}}Slice) Load{{$relAlias}}(ctx context.Context, exe
   if len(os) == 0 {
 	  return nil
 	}
-  
+
   // since we are changing the columns, we need to check if the original columns were set or add the defaults
   sq := dialect.SelectQuery{}
   for _, mod := range mods {
@@ -324,7 +332,7 @@ func (os {{$tAlias.UpSingular}}Slice) Load{{$relAlias}}(ctx context.Context, exe
 	}
 
 	q := os.{{relQueryMethodName $tAlias $relAlias}}(append(
-		mods, 
+		mods,
 		{{range $index, $local := $firstSide.FromColumns -}}
 			{{- $toCol := index $firstTo.Columns (index $firstSide.ToColumns $index) -}}
 			{{- $fromCol := index $firstFrom.Columns $local -}}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Handle table names quoted with backticks in mysql query parser. (thanks @luiscleto)
 - Allow matching columns in type replacements by the `autoincr` property as stated in the docs. (thanks @abdusco)
 - Handle dashes and spaces in generated enum values properly (thanks @abdusco)   
+- Check for nullability when loading relationships from Slices (thanks @felipeparaujo)
 
 ## [v0.38.0] - 2025-06-04
 
--- a/gen/templates/models/table/09_rel_query.go.tpl
+++ b/gen/templates/models/table/09_rel_query.go.tpl
@@ -141,12 +141,15 @@ func (os {{$tAlias.UpSingular}}Slice) {{relQueryMethodName $tAlias $relAlias}}(m
       {{ $column := $.Table.GetColumn $local }}
       {{ $colTyp := $.Types.GetNullable $.CurrentPackage $.Importer $column.Type $column.Nullable }}
 			{{$fromCol := index $firstFrom.Columns $local -}}
-      pk{{$fromCol}} := make(pgtypes.Array[{{$colTyp}}], len(os))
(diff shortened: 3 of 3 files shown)


gen/templates/models/table/10_rel_load.go.tpl
```diff
@@ -239,11 +239,19 @@ func (os {{$tAlias.UpSingular}}Slice) Load{{$relAlias}}(ctx context.Context, exe
...
+
 			o.R.{{$relAlias}} = nil
 		}
 	{{- end}}
 
 	for _, o := range os {
+		if o == nil {
+			continue
+		}
+
```
stephenafamo, line 254: I'm not sure these are necessary. The `All()` method will definitely not return any nil values.

stephenafamo (changes requested):

gen/templates/models/table/10_rel_load.go.tpl
felipeparaujo: these two loops are iterating over the receiver of the method (`os`), not the result from `All()`. I added these checks when I kept running into panics after adding the check in `09_rel_query.go.tpl`

felipeparaujo: #492 explains it with an example:

`Content` has an optional `to-one` relationship to `Sources`
`SourceFiles` has a `to-one` relationship to `Sources`, which implies a `to-many` relationship from the `Sources` side

So suppose we have the following relationship chains:

`content1` -> `source1` -> `sourcefile1`
`content2` -> (nothing)

executing 

```
Contents.Query(
		Preload.Content.Source(
			SelectThenLoad.Source.SourceFiles(),
		),
	).All(context.Background(), db)
```

will panic because these checks I'm updating are trying to `SelectThenLoad` `SourceFiles` on a `SourceSlice` that contain nil elements. IMO it makes sense for this Slice to have nil elements as `Preload.Content.Source` executes a left join, leaving the possibility of some of the results to be nil

stephenafamo: Actually..... I don't think this should happen. Before doing the `ThenLoad`, the slice should be stripped of all nil values.

I'll have to look into this.

stephenafamo (approved):

PR #514, opened 2025-07-25
Remove Generic Type Aliases

daddz:
So this isn't really a bob problem but more a Jetbrains Goland problem.

It doesn't seem to (fully?) support generic type aliases when doing type inference, neither in the current latest release nor in the beta version.

Here are two bug reports related to that:
* https://youtrack.jetbrains.com/issue/GO-17613/Alias-of-generic-type-Cannot-use-alias-as-the-type
* https://youtrack.jetbrains.com/issue/GO-18524/Type-generic-alias-not-recognize-correctly

This leads to problems when working in the IDE and using the `Insert` or `Update` functions mainly since it cannot infer the returned model and as such it breaks auto-completion and displays misleading errors.

I'm not sure when this will be fixed in the IDE but I thought that more people will run into this problem in the meantime.

If you're fine with this solution (for the time being) I would also adapt the sqlite dialect as well - mysql is a bit different and I'm not sure if the alias used there has any impact.

Sidenote: VSCode with the go extension works fine.

### DIFF
--- a/dialect/psql/table.go
+++ b/dialect/psql/table.go
@@ -14,25 +14,18 @@ import (
 	"github.com/stephenafamo/bob/orm"
 )
 
-type (
-	setter[T any]                      = orm.Setter[T, *dialect.InsertQuery, *dialect.UpdateQuery]
-	ormInsertQuery[T any, Tslice ~[]T] = orm.Query[*dialect.InsertQuery, T, Tslice, bob.SliceTransformer[T, Tslice]]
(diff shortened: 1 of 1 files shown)


stephenafamo:
I'm not sure about this.  
While i understand the frustration working like this, I'm not sure Bob should be making changes due to a Goland _bug_.

I assume JetBrains will fix this soon, so perhaps you're better off using your fork until it is fixed.

Issue #517, opened 2025-07-29
Mysql-gen uppercasing some column names in generated code inconsistently


luis-mozn:
New 0.39.0 release broke one of our queries due to a column names 'is_ex' (boolean).

While the Row type has a property `IsEx`, the generated query method scans it using the property in upper case:
```
row.ScheduleScanByIndex(1, &t.ISEX)
```

So the generated code does not compile

Only workaround I found for now seems to be explicit aliasing but this prevents us from using things like `select *`.

luis-mozn:
@stephenafamo I believe the issue was introduced in https://github.com/stephenafamo/bob/commit/18b3c929e8f800cc9e08cfb6543cd59f4a1b5979. This changed the template to use $query.Columns.WithNames but calls titleCase on $col.Name. 'WithNames' should already be title cased, I think

PR #518, opened 2025-07-29
fix(gen): Fix issue due to duplicate titlecasing column names

luiscleto:
Should fix https://github.com/stephenafamo/bob/issues/517 (tested locally on my problematic queries, having issues running mysql gen suite + not sure the best way to add a testcase here).

This makes the $col.Name consistent in the struct declaration and scanner function. Title case is already applied once so applying it a second time can break some column names (eg: is_ex -> IsEx -> ISEX)


luiscleto:
For more details, this seems to happen when using https://github.com/aarondl/strmangle's TitleCase when all lowercase letters are consonants. Whether this can be considered a bug in strmangle (should consider uppercase vowels in each word) I'm not certain

Issue #519, opened 2025-07-29
MySQL setter template generates invalid expression with opt/null and database/sql type systems


tak848:
When using `type_system: github.com/aarondl/opt/null` or `database/sql`, the MySQL setter template generates invalid Go expressions that fail to compile.

Version: v0.39.0

## Problem

The MySQL template generates:
```go
if !s.ID != nil {
    return mysql.Raw("DEFAULT").WriteSQL(ctx, w, d, start)
}
```

This is interpreted as `(!s.ID) != nil` due to operator precedence, causing a compilation error.

## Root Cause

The MySQL template uses `if !{{$.Types.IsOptionalValid ...}}` pattern, which works for the default type system (`github.com/aarondl/opt`) where ValidExpr is `"SRC.IsValue()"`, but fails for:
- `github.com/aarondl/opt/null`: ValidExpr is `"SRC != nil"`
- `database/sql`: ValidExpr is `"SRC != nil"`

## Expected Behavior

The generated code should compile correctly for all supported type systems.

## Configuration to Reproduce

```yaml
mysql:
  dsn: "root:password@tcp(localhost:3306)/test"
  output: "./models"
type_system: github.com/aarondl/opt/null
```

tak848:
I've submitted a PR to fix this: https://github.com/stephenafamo/bob/pull/520

PR #520, opened 2025-07-29
fix(gen): Fix MySQL setter template invalid expression with pointer-based type systems

tak848:
Fixes https://github.com/stephenafamo/bob/issues/519

This PR fixes a compilation error in MySQL setter generation when using `github.com/aarondl/opt/null` or `database/sql` type systems.

The previous pattern generated invalid expressions with pointer-based type systems:
- `github.com/aarondl/opt`: `if !s.ID.IsValue()` ✓ (works)
- `github.com/aarondl/opt/null`: `if !s.ID != nil` ✗ (compile error) 
- `database/sql`: `if !s.ID != nil` ✗ (compile error)

The `!s.ID != nil` pattern was interpreted as `(!s.ID) != nil` causing compilation errors.

This change wraps the ValidExpr with parentheses to fix the operator precedence issue while maintaining the existing template structure.


stephenafamo:
I think the better way to solve the problem will be to wrap the `ValidExpr` with parenthesis.

tak848:
@stephenafamo Thank you for the feedback! I've updated it to use parentheses instead - that's indeed a simpler fix.

The change now wraps the ValidExpr with parentheses:
- `github.com/aarondl/opt`: `if !(s.ID.IsValue())` ✓
- `github.com/aarondl/opt/null`: `if !(s.ID != nil)` ✓ 
- `database/sql`: `if !(s.ID != nil)` ✓

Issue #527, opened 2025-08-06
types.JSON incompatible with PostgreSQL JSON/JSONB columns - returns []byte instead of string


Maxitosh:
Hey! First off, love Bob ORM - it's been great to work with. 
I ran into something curious with PostgreSQL and JSON columns and wanted to check if I'm missing something obvious.

When I use `types.JSON` with PostgreSQL, I get this error: `ERROR: invalid input syntax for type json (SQLSTATE 22P02)`

It looks like `Value()` returns `[]byte`, which PostgreSQL doesn't seem to like for JSON/JSONB columns. 

I found this similar issue in pgtype: https://github.com/jackc/pgtype/issues/45 - they had the same problem where PostgreSQL's text protocol expects strings instead of byte arrays for JSON.

A few questions:
1. Is this expected behavior? 
2. Am I supposed to handle this differently when using PostgreSQL?
3. Would it make sense for Bob to return `string` from `Value()` for JSON types?

PR #530, opened 2025-08-07
Fix JSON/JSONB type handling with pgx driver

Maxitosh:
## Issue
https://github.com/stephenafamo/bob/issues/527

## Problem
When using Bob's `types.JSON` with pgx through `database/sql` (simple protocol), the JSON data was being incorrectly hex-encoded as bytea, causing PostgreSQL to reject it with: `ERROR: invalid input syntax for type json (SQLSTATE 22P02)`

This occurred because `types.JSON.Value()` was returning `[]byte`, which pgx interprets as bytea data when the destination PostgreSQL type is unknown (as is the case with the simple protocol).

## Solution

Modified `types.JSON.Value()` to return a `string` instead of `[]byte`. This aligns with how pgx's own `pgtype.JSON` handles this edge case and ensures proper JSON transmission to PostgreSQL.

## References

- Related pgx issue: https://github.com/jackc/pgtype/issues/45
- pgx implementation: https://github.com/jackc/pgx/blob/master/values.go#L40-L63

### DIFF
--- a/types/json.go
+++ b/types/json.go
@@ -27,7 +27,15 @@ type JSON[T any] struct {
 
 // Value implements the driver Valuer interface.
 func (j JSON[T]) Value() (driver.Value, error) {
-	return json.Marshal(j)
+	data, err := json.Marshal(j)
+	if err != nil {
+		return nil, err
+	}
+	// Return string instead of []byte to avoid pgx treating it as bytea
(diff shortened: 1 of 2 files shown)


stephenafamo:
Kindly add a note in `CHANGELOG.md` about this (with attribution)

PR #555, opened 2025-08-28
SQLite INTEGER PRIMARY KEY as int64

absolutejam:
Hey again,

In the SQLite docs, it states:

> In SQLite, a column with type INTEGER PRIMARY KEY is an alias for the [ROWID](https://www.sqlite.org/lang_createtable.html#rowid) (except in [WITHOUT ROWID](https://www.sqlite.org/withoutrowid.html) tables) which is always a 64-bit signed integer.

But currently the parser sees `INTEGER` and generates `int32` (Which is sensible, but SQLite isn't 😆).

Hopefully this approach is okay - or do you want some special parsing so that we don't obfuscate the actual DB source type?

Also, I'm getting the following failures but I'm not quite sure where they're coming from - are these generated in-memory?

```
-----------------
models/autoinctest.bob.go:378:7: cannot use omit.From(tag1.ID) (value of struct type omit.Val[int32]) as omit.Val[int64] value in struct literal
 375|
 376| func attachAutoinctestIDTag0(ctx context.Context, exec bob.Executor, count int, autoinctest0 *Autoinctest, tag1 *Tag) (*Autoinctest, error) {
 377|   setter := &AutoinctestSetter{
•378|           ID: omit.From(tag1.ID),
 379|   }
 380|
 381|   err := autoinctest0.Update(ctx, exec, setter)
 382|   if err != nil {
-----------------

-----------------
models/autoinctest.bob.go:538:17: invalid operation: o.ID == rel.ID (mismatched types int64 and int32)
 535|
 536|           for _, rel := range tags {
 537|
•538|                   if !(o.ID == rel.ID) {
 539|                           continue
 540|                   }
 541|
 542|                   rel.R.IDAutoinctest = o
-----------------

-----------------
models/tags.bob.go:404:20: cannot use omit.From(tag0.ID) (value of struct type omit.Val[int32]) as omit.Val[int64] value in assignment
 401| }
 402|
 403| func insertTagIDAutoinctest0(ctx context.Context, exec bob.Executor, autoinctest1 *AutoinctestSetter, tag0 *Tag) (*Autoinctest, error) {
•404|   autoinctest1.ID = omit.From(tag0.ID)
 405|
 406|   ret, err := Autoinctests.Insert(autoinctest1).One(ctx, exec)
 407|   if err != nil {
 408|           return ret, fmt.Errorf("insertTagIDAutoinctest0: %w", err)
-----------------

-----------------
models/tags.bob.go:416:7: cannot use omit.From(tag0.ID) (value of struct type omit.Val[int32]) as omit.Val[int64] value in struct literal
 413|
 414| func attachTagIDAutoinctest0(ctx context.Context, exec bob.Executor, count int, autoinctest1 *Autoinctest, tag0 *Tag) (*Autoinctest, error) {
 415|   setter := &AutoinctestSetter{
•416|           ID: omit.From(tag0.ID),
 417|   }
 418|
 419|   err := autoinctest1.Update(ctx, exec, setter)
 420|   if err != nil {
-----------------

-----------------
models/tags.bob.go:660:17: invalid operation: o.ID == rel.ID (mismatched types int32 and int64)
 657|
 658|           for _, rel := range autoinctests {
 659|
•660|                   if !(o.ID == rel.ID) {
 661|                           continue
 662|                   }
 663|
 664|                   rel.R.IDTag = o
-----------------
```

### DIFF
--- a/gen/bobgen-sqlite/driver/exclude-tables.golden.json
+++ b/gen/bobgen-sqlite/driver/exclude-tables.golden.json
@@ -1,3281 +1,3179 @@
 {
-	"tables": [
-		{
-			"key": "autoinckeywordtest",
-			"schema": "",
-			"name": "autoinckeywordtest",
-			"columns": [
-				{
-					"name": "id",
-					"db_type": "INTEGER",
-					"default": "auto_increment",
-					"comment": "",
-					"nullable": false,
-					"generated": false,
-					"autoincr": false,
-					"domain_name": "",
-					"type": "int32",
-					"type_limits": null
-				},
-				{
-					"name": "user_id",
-					"db_type": "INT",
-					"default": "",
-					"comment": "",
-					"nullable": false,
-					"generated": false,
-					"autoincr": false,
-					"domain_name": "",
-					"type": "int32",
-					"type_limits": null
-				},
-				{
-					"name": "sponsor_id",
-					"db_type": "INT",
-					"default": "NULL",
-					"comment": "",
-					"nullable": true,
-					"generated": false,
-					"autoincr": false,
-					"domain_name": "",
-					"type": "int32",
-					"type_limits": null
-				},
-				{
-					"name": "something",
-					"db_type": "TEXT",
-					"default": "NULL",
-					"comment": "",
-					"nullable": true,
-					"generated": false,
-					"autoincr": false,
-					"domain_name": "",
-					"type": "string",
-					"type_limits": null
-				},
-				{
-					"name": "another",
-					"db_type": "TEXT",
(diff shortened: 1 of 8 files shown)


stephenafamo:
> Also, I'm getting the following failures but I'm not quite sure where they're coming from - are these generated in-memory?

The tests run some generation to check if it is working as expected.

In this case, I think the mismatch is because any foreign keys are still generated as `int32` which causes the mismatched type to `int64`

stephenafamo:
I think the better approach is to simply generate all integers in sqlite as `int64`, to better match how SQLite works.

Issue #558, opened 2025-08-29
[Queries] [MySQL] New Uint64 type breaks for BIGINT UNSIGNED with interpolateParams=true


luis-mozn:
Opening an issue as I'm not sure if I might be doing something wrong but, with v0.40.2 I tried enabling interpolateParams in our DSN using https://github.com/go-sql-driver/mysql.
This produced great performance improvements but broke some bob sql queries which now fail with

```
        	            	sql: Scan error on column index 0, name "id": unsupported scan type for Uint64: uint64
```

It seems this changed how the mysql driver handles type conversions and broke scanning due to uint64 not being supported.

Any objections to simply supporting uint64 in Uint64.Scan?
If not I can prepare a PR

PR #559, opened 2025-08-29
fix: support scanning Uint64 from uint64 types

luiscleto:
Should fix https://github.com/stephenafamo/bob/issues/558: failure to scan uint64s when interpolateParams=true


stephenafamo:
Add a note in CHANGELOG please

Issue #585, opened 2025-10-14
Many-to-Many Loader Skipping Related Rows


TylerJGabb:
BobGen version: v0.41.1
Dialect: Postgres

Schema
```sql
CREATE TABLE users (
  id          SERIAL PRIMARY KEY,
  name        text NOT NULL,
  deleted_at  timestamptz NULL
);

CREATE TABLE groups (
  id          SERIAL PRIMARY KEY,
  name        text NOT NULL,
  deleted_at  timestamptz NULL
);

CREATE TABLE user_groups (
  group_id    int NOT NULL REFERENCES groups(id) ON DELETE CASCADE,
  user_id     int NOT NULL REFERENCES users(id) ON DELETE CASCADE,
  PRIMARY KEY (group_id, user_id)
);
```

Steps to Reproduce

Insert two users and one group, then link both users to the group via user_groups.
Generate models with BobGen and run code equivalent to:

```go
groups, _ := models.Groups.Query(
    models.SelectWhere.Groups.DeletedAt.IsNull(),
).All(ctx, db)

_ = groups.LoadUsers(ctx, db)

fmt.Println(len(groups[0].R.Users)) // prints 0
```

Expectation: the group’s .R.Users slice contains both related users.
Actual: .R.Users is empty.
What’s wrong in the generated loader

```go
//orm/models/groups.bob.go, GroupSlice.LoadUsers:

	for _, o := range os {
		for i, rel := range users {
			if o.ID == IDSlice[i] { // should be `!=`
				continue
			}

			rel.R.Groups = append(rel.R.Groups, o)

			o.R.Users = append(o.R.Users, rel)
		}
	}
```
IDSlice is populated with the owning group id for each row (via ScheduleScanByName("related_groups.ID", ...)). The intention was to skip users whose related_groups.ID doesn’t match the current group. Instead, the condition skips the matching ones. Reversing the check to if group.ID != IDSlice[i] { continue } restores the expected behavior.

Impact

Any Many-to-Many loader generated by BobGen that uses this pattern will never hydrate the relation because it discards matches. The workaround is to patch the generated file after each run or correct the template in BobGen.

Suggested fix

Adjust the generated template so the loop condition compares inequality (!=) when deciding whether to continue. I think I found the associated template https://github.com/stephenafamo/bob/blob/66620ae1349979b7564c3d71e6316cd8cc1494c6/gen/templates/loaders/table/110_loaders.go.tpl#L276-L282

PR #586, opened 2025-10-14
fix: #585 Many-to-Many Loader Skipping Related Rows

TylerJGabb:
# Fix: #585 Many-to-Many Loader Skipping Related Rows

## Problem

Generated many-to-many relationship loaders were skipping all related rows instead of hydrating them, causing empty relationship collections.

**Impact**: `groups.LoadUsers(ctx, db)` would result in `groups[0].R.Users` being empty even when relationships existed.

## Root Cause

In `gen/templates/loaders/table/110_loaders.go.tpl` (line ~382), the template generated inverted comparison logic:

```go
// INCORRECT - Generated code was doing this:
if o.ID == IDSlice[i] {  // This skips MATCHING records
    continue
}
```

The condition should skip **non-matching** records, not matching ones.

## Solution

Wrapped the comparison expression in `!(...)` to invert the logic:

```diff
{{- with $.Types.GetCompareExpr $.CurrentPackage $.Importer $fromColDef.Type $fromColDef.Nullable $toColDef.Nullable -}}
-   if {{replace "AAA" $fromColGet . | replace "BBB" $toColGet}} {
+   if !({{replace "AAA" $fromColGet . | replace "BBB" $toColGet}}) {
        continue
    }
{{- end}}
```

**Result**: Generated code now correctly processes matching records:
```go
if !(o.ID == IDSlice[i]) {  // Skip non-matching records
    continue
}
```

## Files Changed
- `gen/templates/loaders/table/110_loaders.go.tpl`

This minimal fix resolves the issue for all many-to-many relationships without breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed a bug where the generated loaders for many-to-many relationships were skipping related rows.
+- Fixed a bug where the generated loaders for many-to-many relationships were skipping related rows. (thanks @TylerJGabb)
 
 ### Added
 


TylerJGabb:
@stephenafamo done, is this what you meant?

PR #598, opened 2025-12-13
Add Relationship Count Loading Support + MySQL Insert Loader Support

jacobmolby:
### Summary

This PR adds a new `counts` plugin that generates code for efficiently loading relationship counts, and allows MySQL insert queries to properly execute loaders.

#### New Counts Plugin

Adds the ability to load counts of to-many relationships without fetching the full related data. This is useful for displaying "X has 5 posts" without loading all posts.

**PreloadCount** - Loads counts via correlated subquery in a single query:
```go
pilots, err := models.Pilots(
    models.PreloadCount.Pilot.Jets(
        models.SelectWhere.Jet.Active.EQ(true), // optional filtering
    ),
).All(ctx, db)

// SQL: SELECT *, (SELECT count(*) FROM jets WHERE jets.pilot_id = pilots.id AND jets.active = true) AS "__count_Jets" FROM pilots

fmt.Printf("Pilot has %d jets\n", *pilots[0].C.Jets)
```

**ThenLoadCount** - Loads counts via separate query after main query:
```go
pilots, err := models.Pilots(
    models.ThenLoadCount.Pilot.Jets(),
).All(ctx, db)
```

**InsertThenLoadCount** - Loads counts after insert:
```go
pilot, err := models.Pilots.Insert(
    &models.PilotSetter{Name: omit.From("John")},
    models.InsertThenLoadCount.Pilot.Jets(),
).One(ctx, db)
```

_The InsertThenLoadCount could easily be expanded to UpdateThenLoadCount as well if we want to. I didn't generate it since MySQL update queries doesn't support returning (.One() or All()). However I can make that change as well if want to._

Models now include a `C` struct with `*int64` pointers for each to-many relationship count.

#### MySQL Insert Loader Support

Added support `InsertThenLoad` and `InsertThenLoadCount` for MySQL. Previously the issue was that `bob.Exec` in `insertAll` where running the loaders causing issues with `nil`. Now we save and clear loaders before executing the INSERT, then transfer them to the SELECT query that retrieves the inserted row.

### Changes

- **New plugin**: `gen/plugins/counts.go`
- **New templates**: `gen/templates/counts/`
- **Updated**: `gen/templates.go`, `gen/plugins/plugins.go` - Register counts plugin
- **Updated**: `gen/templates/models/table/001_types.go.tpl` - Add `C` struct block
- **Fixed**: `dialect/mysql/table.go` - Properly handle loaders in insert flow

### Configuration

```yaml
plugins:
  counts:
    disabled: false  # Enable/disable the counts plugin
```

### Breaking Changes

None. The `C` struct is added to models but doesn't affect existing functionality.

---

**I had difficulties figuring out how the generated queries are tested against a DB? Can you point me in the right direction for this, since I would like to include some actual query tests.**

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -262,10 +262,10 @@ func buildPkgMap(outputs []*Output) (map[string]string, error) {
 }
 
 func tableNames[C, I any](tables drivers.Tables[C, I]) []string {
-	result := make([]string, 0, len(tables))
+	result := make([]string, len(tables))
 
-	for _, table := range tables {
-		result = append(result, table.Key)
+	for i, table := range tables {
+		result[i] = table.Key
 	}
 
 	return result


jacobmolby:
I see. But shouldn't there be a way of writing tests that only the library uses. I feel that writing db tests that are being generated into the consumers code base is very brittle and extremely tedious to write since it needs to account for all possible configurations there are.

Wouldn't it be better if there was some way to just test against a known DB schema (ie. `schema.sql`)? This would also allow for quite easy testing of specific queries using the generated example/test code.

jacobmolby:
@stephenafamo I've tried to add tests specific to each dialect and its corresponding schema. It's run by the library, so the code generation and the model methods/mods can be tested without adding it to the consumers codebase.

I've tried to test a little bit of the everything, however it can definitely be added upon. Especially inserts, updates, deletes and more complex selects can be improved.

Changes: https://github.com/stephenafamo/bob/pull/598/changes/9a1a32a34148573a56228d26e878e165810071a7..e4b254763ff024c7e62f7bae9c27d4628b0648bc

gen/drivers/tables.go
```diff
@@ -22,6 +22,16 @@ func (tables Tables[C, I]) Get(name string) Table[C, I] {
 	panic(fmt.Sprintf("could not find table name: %s", name))
 }
 
+// Has checks if a table with the given name exists.
+func (tables Tables[C, I]) Has(name string) bool {
```
stephenafamo: If we're only using this in the tests, perhaps it can either


* Be converted to a stanalone function in the test file.
* Be moved to the tests file
* Be unexported

Every exported method is something that cannot be changed without a new version (breaking changes).   
Doesn't matter too much yet since Bob is not yet at v1.0, but it would matter more later.

gen/gen.go
```diff
@@ -259,3 +260,13 @@ func buildPkgMap(outputs []*Output) (map[string]string, error) {
 
 	return pkgMap, nil
 }
+
+func tableNames[C, I any](tables drivers.Tables[C, I]) []string {
+	result := make([]string, 0, len(tables))
+
+	for _, table := range tables {
+		result = append(result, table.Key)
+	}
```
stephenafamo:
```suggestion
	result := make([]string, len(tables))

	for i, table := range tables {
		result[i] = table.Key
	}
```

Minor change. Makes no real difference

stephenafamo (commented):
Looks good. Just one question:
* Is it possible to merge the `bobfactory_runtime` tests? Or do they have to be independent?

PR #603, opened 2026-01-06
feat: Add support for older versions of PostgreSQL

wjhdec:
Test Version： Postgresql 12

Before retrieving `index` information, first check if the `pg_index` table has the `indnullsnotdistinct` column: 

- If the column exists, read its value and write it to `nulls_not_distinct`;
- If the column does not exist, handle it using PostgreSQL's default value of `false`.


stephenafamo:
Nice!

Final request, add a note to CHANGELOG.md with your changes, ending with `@thanks @wjhdec`

PR #604, opened 2026-01-06
feat: Add field comments to tags

wjhdec:
Add `comment_tag` to the configuration: - If not set, processing will follow the original logic; - If this configuration is non-empty, the comment of the current field will be added to the generated model (including Setter methods).

Corresponding replacements will be made to the generated tag content:

- Replace the `` ` `` symbol with `'`
- Replace `"` with `\"`
- Replace `\n\t\r\v` symbols with spaces
- Remove leading and trailing spaces

### Testing

Configuration:

```yaml
sql:
  dialect: psql
  pattern: "schema/*.sql"
  
comment_tag: doc
```

schema/test.sql

```sql
create table test_table (
	id varchar primary key,
	demo varchar NULL
);
comment on column test_table.demo is 'Test Comment
With "some words" or `else`
Line1
Line2';
```

After execution, the generated code is as follows:

```go
// TestTable is an object representing the database table.
type TestTable struct {
	ID string `db:"id,pk" `
	// Test Comment
	// With "some words" or `else`
	// Line1
	// Line2
	Demo null.Val[string] `db:"demo"  doc:"Test Comment With \"some words\" or 'else' Line1 Line2"`
}

// TestTableSetter is used for insert/upsert/update operations
// All values are optional, and do not have to be set
// Generated columns are not included
type TestTableSetter struct {
	ID   omit.Val[string]     `db:"id,pk" `
	Demo omitnull.Val[string] `db:"demo"  doc:"Test Comment With \"some words\" or 'else' Line1 Line2"`
}
```


stephenafamo:
Why do you prefer to also have the comment in the struct tag?

I'm curious about the usefulness

stephenafamo:
I am reluctant to add this to Bob. 

For your use case, I would suggest instead creating a personal plugin that creates specific structs from the DB for huma. You can add the doc tag, and perhaps functions to convert from the models to the huma structs.

stephenafamo:
If the plugin adds enough huma-specific helpers, I'll be happy to add it to the docs for anyone else looking to use Bob with Huma

stephenafamo:
> Okay, this might be useful when you need to use reflection to get comments, not just in Huma.

While that may be true, one big benefit of code generation is to avoid using comments.  

If a user really needs to access comments, the `dbinfo` plugin generates code that includes the comments which can be retrieved without any reflection.

PR #614, opened 2026-01-21
gen mysql driver includes a check for mariaDB compatibility

dumdev25:
mysql has extensions column for  information_schema.statistics table, which doesn't exist for mariaDB.
This PR includes a check to default that column to null, making it compatible with mariaDB


dumdev25:
Thank you so much for your quick feedback.

stephenafamo:
Kindly fix the lint issue and I'll be happy to merge

PR #615, opened 2026-01-22
Return error if SET clause is empty

### LINKED ISSUE #608: Empty Setter struct generates incorrect SQL when passed to Update
When using a `*Setter` struct in an `Update` method the generated SQL will contain an empty `SET` clause if the `*Setter` struct didn't contain any updated values. This is understandable as there is no SQL statement to put here which is both valid and useful. However I think it would be better if the `Update` method simply did nothing and returned `nil` in this situation. No update was required, so this is a correct response.

Xaeroxe:
Closes #608 

After conducting an exhaustive review of the syntax for MySQL, PostgreSQL, and SQLite, I have determined that all contexts in which a `SET` clause appears require the user to specify at least one expression.

Here are some sources to review if you wish to double check this claim:

- https://www.postgresql.org/docs/current/sql-update.html
- https://www.postgresql.org/docs/current/sql-insert.html
- https://www.sqlite.org/lang_update.html
- https://www.sqlite.org/lang_insert.html
- https://dev.mysql.com/doc/refman/8.4/en/update.html
- https://dev.mysql.com/doc/refman/8.4/en/insert.html

To that end, I am proposing that we throw an error if an empty `SET` clause is written to SQL in any context, for any reason.

### DIFF
--- a/clause/set.go
+++ b/clause/set.go
@@ -8,7 +8,7 @@ import (
 	"github.com/stephenafamo/bob"
 )
 
-var ErrUpdateHasNoSetExpressions = errors.New("SET clause must have at least one assignment expression")
+var ErrEmptySetExpression = errors.New("SET clause must have at least one assignment expression")
 
 type Set struct {
 	Set []any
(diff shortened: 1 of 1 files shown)


stephenafamo:
Fix the lint issues please

Issue #618, opened 2026-01-27
Support sub query expressions (ANY, SOME, ALL, EXISTS) in PostgreSQL and MySQL


manhrev:
I can't find a way to express sub query expressions, for examples:
```
# ANY / SOME / ALL subquery
...
WHERE book_type = ANY (?)

# EXISTS subquery
...
WHERE EXISTS(
   SELECT 1 FROM ...
)
```

Currenty, I workaround this by using `F()` but I don't think this is the corect approach:
```
sm.Where(
   psql.F("EXISTS", existQuery),
)
```

I think we should add dedicated starters for those expressions, such as `Exists(...Expression)`, `Any(...Expression)`, etc. Both [PostgreSQL](https://www.postgresql.org/docs/current/functions-subquery.html) and [MySQL](https://dev.mysql.com/doc/refman/8.4/en/subqueries.html) support those operator, while SQLite supports `EXISTS` only.
Actually, `IN` and `NOT IN` belong to the same category of operators and are already supported through operator chaining but other operators that I mention can't be implemented like that.

Bonus: I also can’t find a way to express the unary minus operator (-). Should this be added as a starter as well?

stephenafamo:
I think `Exists` and `UnaryMinus` can be starters, however, `ANY`, `SOME` and `ALL` should be included as chain methods specific to Postgres (similar to `BetweenSymmetric`)

stephenafamo:
> but there's currently no ways to implement it

Actually there is. This is done by returning the base interface `bob.Expression` instead of the dialect specific `Expression` type. An example of this is the chained `As` method.

PR #622, opened 2026-01-30
Add EXISTS, ALL, SOME, ANY expressions

manhrev:
Implements the feature requested in #618

### DIFF
--- a/dialect/mysql/starters.go
+++ b/dialect/mysql/starters.go
@@ -125,12 +125,6 @@ func Any(exp bob.Expression) bob.Expression {
 	return bmod.Any(exp)
 }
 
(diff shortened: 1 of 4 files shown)


stephenafamo:
Thank you. Great work 🎉 

Just a few comments:

1. Update `ANY`, `SOME`, `ALL` to return `bob.Expression` so that they are not chainable.
2. Add a note in `CHANGELOG.md`, ending with `(thanks @manhrev)`

manhrev:
Updated!
I also added `ANY`, `ALL`, and `SOME` for the MySQL dialect.

manhrev:
Yeah, I think people generally prefer ANY over SOME, so keeping only ANY is fine

PR #625, opened 2026-02-01
feat: add Unqualified method to return columns without table alias/pr…

atzedus:
### Summary

This PR adds a new `Unqualified()` method to generated column structures that returns columns without table alias/prefix. This is useful when you need to reference column names without table qualification, for example in `INSERT ... ON CONFLICT` clauses or when building dynamic queries.

### Changes

- Added `Unqualified()` method to generated `*Columns` structures
- Modified `buildXColumns()` function to conditionally call `WithParent()` only when `tableName` is not empty
- Renamed parameter from `alias` to `tableName` for better clarity

### Example

Before this change, columns were always qualified with table name:
```go
Orders.Columns.ID  // generates: "orders"."id"
```

Now you can get unqualified column names:
```go
Orders.Columns.Unqualified().ID  // generates: "id"
```

### Use Case

This is particularly useful for `ON CONFLICT` clauses where you need column names without table prefix:
```go
psql.Insert(...).
    OnConflict(Orders.Columns.Unqualified().ID).
    DoUpdate(...)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `Unqualified()` method to generated column structures that returns columns without table alias/prefix. (thanks @atzedus)
 - Added `PreloadCount` and `ThenLoadCount` to generate code for preloading and then loading counts for relationships. (thanks @jacobmolby)
 - MySQL support for insert queries executing loaders (e.g., `InsertThenLoad`, `InsertThenLoadCount`). (thanks @jacobmolby)
 - Added overwritable hooks that are run before the exec or scanning test of generated queries. This allows seeding data before the test runs.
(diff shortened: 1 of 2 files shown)


atzedus:
> If `Unqualified` is now the same as calling `AliasedAs("")`, do we really need an additional method?
> 
> I try to avoid adding new methods so there are less avenues for breaking changes

You're right that `Unqualified()` is functionally equivalent to `AliasedAs("")`. However, I believe there are good reasons to keep it:

1. Semantic clarity

`Unqualified()` clearly expresses intent, while `AliasedAs("")` looks like a magic value or potential bug. When reading code:

```go
// Clear intent - "give me columns without table prefix"
Orders.Columns.Unqualified()

// Confusing - "why empty string? Is this a mistake?"
Orders.Columns.AliasedAs("")
```

2. It affects `ColumnsExpr` methods, not just individual column expressions

The key difference is that `Unqualified()` returns columns where `ColumnsExpr.parent` is empty. This affects methods like `Only()`, `Except()`, `WithPrefix()` that operate on the embedded `ColumnsExpr`:

```go
// With Unqualified() - generates: "id", "name" (no table prefix)
Orders.Columns.Unqualified().Only("id", "name")

// The parent slice is empty, so WriteSQL produces clean column list
// Useful for INSERT column lists, ON CONFLICT targets, etc.
```

3. Discoverability

Users looking for "how to get columns without table name" will find `Unqualified()` through autocomplete. They won't think to try `AliasedAs("")`.

That said, I understand your concern about API surface. If you prefer, we could:
- Document `AliasedAs("")` as the way to get unqualified columns
- Or keep `Unqualified()` as a convenience method that's unlikely to need breaking changes

What do you think?

stephenafamo (commented):
Just a little nitpick

PR #626, opened 2026-02-03
fix: Reuse parents when creating children

### LINKED ISSUE #510: ChildMods.WithoutParent does nothing, factory generates a new parent entity for every associated child
Factory generates a new parent entity for every associated child, causing many parent records to be created. To illustrate the issue, imagine we have this schema:

```sql
CREATE TABLE orders
(
    id SERIAL PRIMARY KEY
);

CREATE TABLE order_items
(
    id       SERIAL PRIMARY KEY,
    order_id INTEGER NOT NULL REFERENCES orders (id)
);
```

and this config:

```yaml
psql:
  driver: github.com/jackc/pgx/v5/stdlib
  type_system: database/sql
  output: test/db
  pkgname: dbtest
  schemas:
    - public
  uuid_pkg: gofrs
```

When I try to create an order with 2 order items, the factory generates a new order for every child item, so 3 orders are created in total. 

```go

func TestChild(t *testing.T) {
	f := factory.New()

	// orders should be created with 2 items
	f.AddBaseOrderMod(
		factory.OrderMods.WithNewOrderItems(2),
	)
	// doesn't do anything
	f.AddBaseOrderItemMod(
		factory.OrderItemMods.WithoutOrder(),
	)

	db := must(sql.Open("postgres", "dbname=bob sslmode=disable"))

	bobDB := bob.NewDB(db)
	ctx := context.Background()
	withRollback(ctx, bobDB, func(tx bob.Transaction) error {
		_, err := f.NewOrder().Create(ctx, tx)
		if err != nil {
			return err
		}
		total, err := models.Orders.Query().Count(ctx, tx)
		if err != nil {
			return err
		}

		t.Log(total) // prints 3, expected 1

		return nil
	})
}

func must[T any](v T, err error) T {
	if err != nil {
		panic(err)
	}
	return v
}

func withRollback(ctx context.Context, db bob.DB, fn func(tx bob.Transaction) error) {
	tx := must(db.BeginTx(ctx, nil))
	defer tx.Rollback(ctx)

	if err := fn(tx); err != nil {
		panic(err)
	}
}
```

This seems to be caused by this bit. 

https://github.com/stephenafamo/bob/blob/08202428a120d26097136c41841fc636afa236a1/gen/templates/factory/table/03_create.go.tpl#L117-L118

Adding `.OrderItemMods.WithoutOrder()` doesn't do anything, because it results in this code, which always creates a parent even though we told it not to.

```go
func (m orderItemMods) WithoutOrder() OrderItemMod {
	return OrderItemModFunc(func(ctx context.Context, o *OrderItemTemplate) {
		o.r.Order = nil
	})
}

func (o *OrderItemTemplate) Create(ctx context.Context, exec bob.Executor) (*models.OrderItem, error) {
	var err error
	opt := o.BuildSetter()
	ensureCreatableOrderItem(opt)

	if o.r.Order == nil {
		OrderItemMods.WithNewOrder().Apply(ctx, o)
	}
```

abdusco:
Closes: #510

### DIFF
--- a/gen/templates/factory/bobfactory_context.bob.go.tpl
+++ b/gen/templates/factory/bobfactory_context.bob.go.tpl
@@ -14,6 +14,8 @@ var (
       {{- end}}
 
     {{end}}
+
+    modelsInCreationCtx = newContextual[map[string]any]("modelsInCreation")
 )
 
 // Contextual is a convienience wrapper around context.WithValue and context.Value
@@ -33,4 +35,3 @@ func (k contextual[V]) Value(ctx context.Context) (V, bool) {
   v, ok := ctx.Value(k.key).(V)
   return v, ok
 }
-
--- a/gen/templates/factory/bobfactory_main.bob_test.go.tpl
+++ b/gen/templates/factory/bobfactory_main.bob_test.go.tpl
@@ -1,5 +1,17 @@
+{{- $needsModels := false -}}
+{{- range $table := .Tables -}}
+  {{- range $rel := $.Relationships.Get $table.Key -}}
+    {{- $bridgeRels := $.Tables.NeededBridgeRels $rel -}}
+    {{- if and $rel.IsToMany (ne $rel.Foreign $table.Key) (not $bridgeRels) (eq (len $rel.Sides) 1) -}}
+      {{- $needsModels = true -}}
+    {{- end -}}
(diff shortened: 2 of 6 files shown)


stephenafamo:
Add a note to CHANGELOG.md

PR #647, opened 2026-04-10
Batch LoadCount for slices

jacobmolby:
## Batch `LoadCount` for slices                                                                                                                         
                                                            
### Problem                                                                                                                                             
                                                                                                                                                        
`<Model>Slice.LoadCount<Rel>()` previously executed one `SELECT COUNT(*)` query per item                                                                
in the slice - an N+1 pattern. For a slice of 100 videos, loading the tag count fired 100
separate queries.                                                                                                                                       
                                                          
### Solution                                                                                                                                            
                                                          
Rewrite the generated slice `LoadCount` methods to execute a **single batch query** that                                                                
fetches counts for all parents at once, then distribute results to each model in Go.
                                                                                                                                                        
The generated SQL looks like:                                                                                                                           
                                                                                                                                                        
```sql                                                                                                                                                  
SELECT fk_col, count(*) as count                          
FROM related_table
WHERE fk_col IN ($1, $2, ...)   -- MySQL/SQLite, single-column FK
GROUP BY fk_col                                                                                                                                         
 ```
Results are scanned into a local struct and distributed via a map lookup. Parents with no                                                               
related rows get count = 0.                               
                                                                                                                                                        
### Code paths                                                
                                                                                                                                                        
#### IN clause - dialect-specific                                                                                                                            
 
- MySQL / SQLite, single-column FK: `col IN ($1, $2, ...)` - plain scalar list                                                                            
- MySQL / SQLite, composite FK: `(col1, col2) IN (($1, $2), ($3, $4), ...)` - row value form, which SQLite supports for multi-column tuples
- PostgreSQL: `unnest(CAST($1 AS type[]))` - a single array argument instead of N placeholders, avoiding prepared-statement parameter limits on large slices                                                                                                                                                  
                                                                                                                                                        
                                 
#### Relationship hops                                                                                                                                       
                                                                                                                                                        
- **Single-hop** (e.g. `Video → Tags` via direct FK):
  `FROM related_table WHERE fk IN (...) GROUP BY fk`
- **Multi-hop** (e.g. `Video → Tags` via `video_tags` join table):
  `FROM join_table INNER JOIN related_table ON ... WHERE join_fk IN (...) GROUP BY join_fk`
  - Static `FromWhere`/`ToWhere` filters are applied per side                                                                                         
                                                                                                                                                        
#### FK arity

- **Single-column FK**: map key is the FK type directly - `map[uint64]int64`
- **Composite FK**: a local `countKey` struct is generated as the map key - `map[countKey]int64`

#### `ThenLoadCount`

No changes needed. `ThenLoadCount` delegates to `LoadCount` via `AfterPreloader` and
automatically benefits from this change.

#### Single-model `LoadCount`

Unchanged - continues to use the typed query method (`.Count(ctx, exec)`), which is
appropriate for the single-item case.

### DIFF
--- a/gen/templates/counts/table/115_counts.go.tpl
+++ b/gen/templates/counts/table/115_counts.go.tpl
@@ -171,12 +171,18 @@ func (os {{$tAlias.UpSingular}}Slice) LoadCount{{$relAlias}}(ctx context.Context
 		if o == nil {
 			continue
 		}
+		{{if eq (len $firstSide.FromColumns) 1 -}}
+		{{$local := index $firstSide.FromColumns 0 -}}
+		{{$fromCol := index $firstFrom.Columns $local -}}
+		PKArgSlice = append(PKArgSlice, {{$.Dialect}}.Arg(o.{{$fromCol}}))
+		{{- else -}}
 		PKArgSlice = append(PKArgSlice, {{$.Dialect}}.ArgGroup(
 			{{- range $index, $local := $firstSide.FromColumns -}}
 			{{- $fromCol := index $firstFrom.Columns $local -}}
 			o.{{$fromCol}},
 			{{- end -}}
 		))
+		{{- end}}
 	}
 	PKArgExpr := {{$.Dialect}}.Group(PKArgSlice...)
 	{{- else}}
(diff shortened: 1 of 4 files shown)


stephenafamo:
Looks good. Update the CHANGELOG.md 🙏🏾

jacobmolby:
> Looks good. Update the CHANGELOG.md 🙏🏾

How should I add it? The original feature is still under "unreleased"? 😄

stephenafamo:
Oh 😅

Issue #39, opened 2023-03-31
Materialized view columns


hiendaovinh:
Hi,
With the given schema: https://github.com/stephenafamo/bob/blob/main/gen/bobgen-psql/driver/testdatabase.sql, I think it's reasonable to expect bob to generate `type_monsters_mv` columns just like `type_monsters_v` but it doesn't. Is it expected behavior?
https://github.com/stephenafamo/bob/blob/main/gen/bobgen-psql/driver/psql.golden.json#L2228-L2238
Thank you.

stephenafamo:
It would be a reasonable expectation, I just haven't gotten to it yet.

Because materialized views are not part of the SQL specification, Postgres does not include information about them in the `information_schema.views` table.  
So querying for details of a materialized view needs a whole new query in the driver.

I'm not sure how soon I'll get to this myself. Of course, I'll be happy to review a PR for this 😁.

hiendaovinh:
I'll be happy to work on this. Let me test `pg_get_viewdef` and `pg_matviews` and see what I can do.

stephenafamo:
If you can implement this [`TableDetails` function](https://github.com/stephenafamo/bob/blob/main/gen/bobgen-psql/driver/psql.go#LL201-L372C2) for materialized views, then the rest would be a breeze.

hiendaovinh:
Yes I'm looking on it.
- [x] column_name: attname
- [ ] column_type (Currently returned actual enum "workday" name instead of "ENUM")
- [ ] udt_schema
- [ ] udt_name
- [ ] array_type
- [ ] domain_name
- [ ] column_default
- [ ] column_comment
- [x] is_nullable: attnotnull
- [x] is_generated: attgenerated
- [x] is_identity: attidentity

Will comeback and update this.

hiendaovinh:
This might also helps: https://github.com/typeorm/typeorm/blob/master/src/driver/postgres/PostgresQueryRunner.ts

stephenafamo:
How's it going? @hiendaovinh

hiendaovinh:
Sorry, there's no further progress yet because I was kind of busy last week. Will come back this week sir.

tlietz:
@hiendaovinh I would like to start working on this if you no longer are. 

[...]

PR #650, opened 2026-04-15
bob-gen psql fetch info about materialized views

PudottaPommin:
I needed to generate schema for materialized views and found this issues #39. I added query to fetch also materialized views metadata.

Tests seem to pass, but I'm not sure if to extended any of golden.jsons.


stephenafamo:
Yes, you'll need to run the tests with the `-overwrite-golden` flag 

Thanks for this 🙏🏾

PudottaPommin:
It does seem that might have make tests pass 🙏

PudottaPommin:
I'm in a bind. When I run tests locally, they pass. Also I'm not sure, why it's also mysql tests failing, when I didn't touch them originally and that part either.

I was running this from Windows, then tried WSL with same result. Works locally, doesn't work in actions. Do you have any idea, what can be core of the problem? Maybe golden files were not properly rewritten?

stephenafamo:
Fixed in #677

stephenafamo:
The issue seems to be that when running on windows, JSON new lines are encoded as `/r/n` while the actions expect it to only be `/n`

PR #656, opened 2026-04-20
feat(psql): add native immutable with paths

jay-babu:
Adds native immutable `With(...)` paths for PostgreSQL `Select`, `ViewQuery`, `Update`, `Delete`, and `Insert` derivations while keeping existing mutable `Apply(...)` behavior unchanged.

The user-facing constructors stay the same:
- `psql.Select(...)`
- `view.Query(...)`
- `psql.Update(...)`
- `psql.Delete(...)`
- `psql.Insert(...)`

The behavior change is opt-in at the call site: callers only see the new immutable behavior and perf gains when they use `With(...)` and keep the returned query.

What changed:
- `Select(...).With(...)` and `ViewQuery.With(...)` now use a native immutable path instead of clone-plus-mutate.
- `Update(...).With(...)`, `Delete(...).With(...)`, and `Insert(...).With(...)` now follow the same pattern.
- common derivation cases use specialized immutable state updates
- unsupported/fallback cases still work by materializing the mutable query internally
- added regression coverage to verify `With(...)` does not mutate the base query and matches direct construction for native and fallback cases

Benchmarks:
```text
BenchmarkBaseQueryApplyMain                               2990 ns/op   2865 B/op    48 allocs/op
BenchmarkBaseQueryImmutableNativeHotPath                  2143 ns/op   1705 B/op    36 allocs/op
BenchmarkViewQueryCountThenPaginateApplyMain            10084 ns/op   7440 B/op   130 allocs/op
BenchmarkViewQueryCountThenPaginateImmutableNativeHotPath 9153 ns/op   6115 B/op   117 allocs/op
BenchmarkUpdateQueryApplyMain                             2433 ns/op   2319 B/op    47 allocs/op
BenchmarkUpdateQueryImmutableNativeHotPath                1966 ns/op   1360 B/op    41 allocs/op
BenchmarkDeleteQueryApplyMain                             1879 ns/op   1956 B/op    33 allocs/op
BenchmarkDeleteQueryImmutableNativeHotPath                1383 ns/op   1027 B/op    26 allocs/op
BenchmarkInsertQueryApplyMain                             1543 ns/op   1576 B/op    24 allocs/op
BenchmarkInsertQueryImmutableNativeHotPath                1157 ns/op    920 B/op    20 allocs/op
```

Verification:
- `go test ./dialect/psql`
- `go test ./dialect/psql ./orm ./mods ./clause`
- `go test ./dialect/psql -run '^$' -bench 'Benchmark(BaseQuery|ViewQueryCountThenPaginate|UpdateQuery|DeleteQuery|InsertQuery)' -benchmem`

Regression coverage added:
- `Select` native path
- `Select` fallback path
- `View.Query(...).With(...)`
- `Update`
- `Delete`
- `Insert` native path
- `Insert` fallback path

### DIFF
--- a/dialect/psql/delete.go
+++ b/dialect/psql/delete.go
@@ -5,15 +5,25 @@ import (
 	"github.com/stephenafamo/bob/dialect/psql/dialect"
 )
 
-func Delete(queryMods ...bob.Mod[*dialect.DeleteQuery]) bob.BaseQuery[*dialect.DeleteQuery] {
+type DeleteQuery struct {
+	bob.BaseQuery[*dialect.DeleteQuery]
+}
+
+func (q DeleteQuery) With(queryMods ...bob.Mod[*dialect.DeleteQuery]) derivedDeleteQuery {
+	return asImmutableDelete(q.BaseQuery).With(queryMods...)
+}
+
+func Delete(queryMods ...bob.Mod[*dialect.DeleteQuery]) DeleteQuery {
 	q := &dialect.DeleteQuery{}
 	for _, mod := range queryMods {
 		mod.Apply(q)
 	}
 
-	return bob.BaseQuery[*dialect.DeleteQuery]{
-		Expression: q,
-		Dialect:    dialect.Dialect,
-		QueryType:  bob.QueryTypeDelete,
+	return DeleteQuery{
+		BaseQuery: bob.BaseQuery[*dialect.DeleteQuery]{
+			Expression: q,
+			Dialect:    dialect.Dialect,
+			QueryType:  bob.QueryTypeDelete,
+		},
 	}
 }
--- a/dialect/psql/immutable_select.go
+++ b/dialect/psql/immutable_select.go
@@ -16,7 +16,7 @@ import (
 	"github.com/stephenafamo/scan"
 )
 
-type ImmutableSelectQuery struct {
+type derivedSelectQuery struct {
 	state immutableSelectState
 }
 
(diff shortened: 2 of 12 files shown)


jay-babu:
@stephenafamo here is the PR based on our discussion with adding immutable. this is backwards compatible so no current users are impacted. in addition, added benchmarks showing the performance benefit. now users can use either `apply` for current behavior and `with` for immutable

stephenafamo:
I don't yet fully understand how the immutable version works, but if according to your benchmark it has less allocations, then I'm tempted to have the immutable version be the only version.

jay-babu:
Right now the immutable path is additive, not a full replacement.

How it works today:
- The public constructors (`Select`, `Update`, `Delete`, `Insert`, `View.Query`) still build the existing mutable query structs.
- `Apply(...)` still mutates exactly as before.
- `With(...)` is the new path. It snapshots the current query state into a compact derived-query representation.
- For the common hot-path mods, `With(...)` updates that derived state directly instead of cloning the whole mutable query and re-running the mod pipeline.
- If `With(...)` sees a mod it does not handle natively yet, it falls back by materializing the mutable query internally, applying the mod there, then converting back into derived state.

So the current perf win is specifically from:
- not cloning the full mutable query
- not re-running generic mutation for the supported derivation cases
- writing SQL directly from the derived state

That is also why existing callers see no change unless they switch from `Apply(...)` to `With(...)`.

If we move to immutable-by-default, the direction would be to invert that design:
- constructors produce the immutable/native representation directly
- derivation APIs operate on that representation only
- mutation becomes either removed or a thin compatibility layer
- the current fallback path shrinks instead of being the main model

I’m doing that follow-up on a separate branch so this PR can stay focused.

PR #657, opened 2026-04-21
refactor(psql): simplify immutable query internals

jay-babu:
## Summary

Relative to `upstream/main`, this PR moves Bob further toward immutable-by-default query building and rewrites the hot `psql` query path so that immutable query derivation stays competitive with or ahead of the current mutable `Apply` path.

What this PR changes:
- makes `bob.BaseQuery` immutable-by-default: `Apply(...)` now returns a derived query instead of mutating in place
- updates `orm.ExecQuery` and `orm.Query` to follow the same immutable semantics
- updates mysql/sqlite table and view wrappers to work with immutable `Apply(...)`
- refactors `psql` to use a single query representation instead of a parallel immutable wrapper model
- adds `Clone()` support and immutable derivation support for the `psql` dialect query structs
- specializes the `psql` writer path to keep immutable derivation fast enough to beat the current upstream baselines on the main benchmark shapes
- simplifies some of the supporting `psql` wrapper and table/view logic after the core rewrite

## Client migration

The main breaking change is semantic:
- `Apply(...)` no longer mutates queries in place
- `Apply(...)` now behaves like `With(...)`: it returns a derived query

So code like this on `upstream/main`:

```go
q := psql.Select(sm.From("users"))
q.Apply(sm.Where(psql.Quote("id").EQ(psql.Arg(1))))
```

must become:

```go
q := psql.Select(sm.From("users"))
q = q.Apply(sm.Where(psql.Quote("id").EQ(psql.Arg(1))))
```

The same applies to pointer-returning wrappers such as `orm.Query`, `orm.ExecQuery`, table queries, and view queries:

```go
q := userTable.Update(...)
q = q.Apply(um.Where(...))
```

If the existing code was already written as:
- `q = q.Apply(...)`, or
- `q2 := q.With(...)`

then no behavior change is required.

Practical migration notes by layer:
- `bob.BaseQuery`: always capture the return value of `Apply(...)`
- `orm.ExecQuery` / `orm.Query`: always capture the return value of `Apply(...)`
- dialect wrappers like `psql.SelectQuery`, `psql.UpdateQuery`, `psql.DeleteQuery`, `psql.InsertQuery`: always capture the return value of `Apply(...)`
- table/view helpers in `psql`, `mysql`, and `sqlite`: same rule, always capture the return value of `Apply(...)`

What does *not* change:
- `With(...)` still returns a derived query
- constructors such as `psql.Select(...)`, `psql.Update(...)`, `psql.Delete(...)`, `psql.Insert(...)`, `view.Query(...)`, and table helpers still exist
- SQL shape and result behavior are intended to remain the same apart from immutable query semantics

## Main areas changed

Core immutable semantics:
- `query.go`
- `orm/query.go`
- `query_immutable_test.go`

`psql` query core and writers:
- `dialect/psql/dialect/clone.go`
- `dialect/psql/dialect/derive.go`
- `dialect/psql/dialect/writer.go`
- `dialect/psql/dialect/select.go`
- `dialect/psql/dialect/update.go`
- `dialect/psql/dialect/delete.go`
- `dialect/psql/dialect/insert.go`

`psql` wrappers and table/view behavior:
- `dialect/psql/select.go`
- `dialect/psql/update.go`
- `dialect/psql/delete.go`
- `dialect/psql/insert.go`
- `dialect/psql/view.go`
- `dialect/psql/table.go`
- `dialect/psql/{sm,um,dm,im}/qm.go`

Compatibility fallout for other dialect wrappers:
- `dialect/mysql/{view,table}.go`
- `dialect/sqlite/{view,table}.go`

## Verification

Ran:
- `go test ./dialect/psql ./dialect/psql/dialect ./orm ./dialect/mysql ./dialect/sqlite`
- `go test ./dialect/psql -run '^$' -bench '^Benchmark(BaseQuery|ViewQueryCountThenPaginate|UpdateQuery|DeleteQuery|InsertQuery)(ApplyMain|ImmutableNativeHotPath)$' -benchmem`

Latest immutable benchmark snapshot on this branch:
- `BenchmarkBaseQueryImmutableNativeHotPath`: `2931 ns/op`, `3149 B/op`, `35 allocs/op`
- `BenchmarkViewQueryCountThenPaginateImmutableNativeHotPath`: `5941 ns/op`, `6316 B/op`, `62 allocs/op`
- `BenchmarkUpdateQueryImmutableNativeHotPath`: `2601 ns/op`, `2575 B/op`, `43 allocs/op`
- `BenchmarkDeleteQueryImmutableNativeHotPath`: `1962 ns/op`, `2156 B/op`, `27 allocs/op`
- `BenchmarkInsertQueryImmutableNativeHotPath`: `1507 ns/op`, `1720 B/op`, `18 allocs/op`

Upstream/main baselines used for comparison:
- `BenchmarkBaseQueryApplyMain`: `3039 ns/op`, `2865 B/op`, `48 allocs/op`
- `BenchmarkViewQueryCountThenPaginateApplyMain`: `10070 ns/op`, `7424 B/op`, `130 allocs/op`
- `BenchmarkUpdateQueryApplyMain`: `3038 ns/op`, `2319 B/op`, `47 allocs/op`
- `BenchmarkDeleteQueryApplyMain`: `1926 ns/op`, `1956 B/op`, `33 allocs/op`
- `BenchmarkInsertQueryApplyMain`: `1473 ns/op`, `1576 B/op`, `24 allocs/op`

## Notes

This PR still adds complexity in the `psql` specialization layer, mainly in:
- `dialect/psql/dialect/writer.go`
- `dialect/psql/dialect/derive.go`

That is the current cost of keeping the immutable path at or above the upstream performance bar on the benchmarked query shapes.


jay-babu:
One migration decision worth calling out separately: we now have two viable paths for client-facing semantics.

1. Remove `Apply(...)` and force all clients onto `With(...)`.
This is the cleaner API because there is only one immutable composition path, and the migration is mechanically understandable: find places that currently call `q.Apply(...)` and switch them to `q = q.With(...)` (or equivalent reassignment). That makes the touched call sites explicit and avoids clients having to remember that `Apply(...)` changed meaning.

2. Keep `Apply(...)` and let clients absorb the semantic change.
This minimizes immediate source edits in some codebases, but it pushes the cognitive cost onto clients because `Apply(...)` no longer behaves the way it did on `upstream/main`. In practice, clients then need to audit every place that previously relied on in-place mutation and make sure the returned query is captured.

My bias is that if we are willing to take a breaking change anyway, removing `Apply(...)` is the cleaner long-term API. If we keep it, we should treat the migration burden as a real client cost and document the exact call sites to audit: `psql.Select`, `view.Query`, table queries, `orm.Query`, and `orm.ExecQuery` derivation chains.

jay-babu:
@stephenafamo Here is the PR migrating everything to be immutable by design

stephenafamo:
1. Is there still any need for `q.With()`?
2. I see a lot of additional changes to `psql` that were not also made to `mysql` and `sqlite`. Is it that only psql is now immutable?

jay-babu:
1. @stephenafamo I can remove With. That was there to introduce it as a nonbreaking change first

2. The base of code is immutable, but I did not focus on going through and making mysql and sqlite immutable by design

jay-babu:
@stephenafamo what will it take to get this merged?

stephenafamo:
Sorry I ignored this for a while... I'm just conflicted because of how big of a breaking change this is.

I'm considering having a side-by-side implementation, i.e. `psql.Select` and `psql.Select2` along with a flag for the code generation to switch between them to allow for less painful migration.

PR #662, opened 2026-04-30
PostgreSQL MERGE parser support

atzedus:
Added MERGE support to the parser.
Performed a minor refactoring (split the code into separate files for INSERT/UPDATE/DELETE/SELECT/MERGE).


gen/bobgen-psql/driver/parser/mods_merge.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func mergeMatchKindToDialectWhenType(kind pg.MergeMatchKind) (string, bool) {
+	switch kind {
+	case pg.MergeMatchKind_MERGE_WHEN_MATCHED:
+		return "dialect.MergeWhenMatched", true
+	case pg.MergeMatchKind_MERGE_WHEN_NOT_MATCHED_BY_SOURCE:
+		return "dialect.MergeWhenNotMatchedBySource", true
+	case pg.MergeMatchKind_MERGE_WHEN_NOT_MATCHED_BY_TARGET:
+		return "dialect.MergeWhenNotMatched", true
```
stephenafamo, line 176: Wouldn't it be more precise to map this to `MergeWhenNotMatchedByTarget`?

gen/bobgen-psql/driver/parser/verify.go
```diff
@@ -41,3 +41,11 @@ func verifyDeleteStatement(stmt *pg.DeleteStmt, _ nodeInfo) error {
...
 	return nil
 }
+
+func verifyMergeStatement(stmt *pg.MergeStmt, _ nodeInfo) error {
+	if stmt == nil {
+		return fmt.Errorf("nil statement")
+	}
+
+	return nil
+}
```
stephenafamo, line 59: Is there nothing to verify?

gen/bobgen-psql/driver/parser/mods_merge.go
atzedus: According to docs

> Note that the SQL standard only defines WHEN MATCHED and WHEN NOT MATCHED (which is defined to mean no matching target row). WHEN NOT MATCHED BY SOURCE is an extension to the SQL standard, as is the option to append BY TARGET to WHEN NOT MATCHED, to make its meaning more explicit.

Therefore, `WHEN NOT MATCHED BY TARGET` is effectively an alias for `WHEN NOT MATCHED`. That said, you might be right that using MergeWhenNotMatchedByTarget makes the intent more explicit.

atzedus: I reverted last commit with comments. pg_query returns the same AST kind for both WHEN NOT MATCHED and WHEN NOT MATCHED BY TARGET, so we cannot reliably preserve the original wording. Mapping to MergeWhenNotMatched keeps identical semantics and generates SQL that still works on PG 15-16.

gen/bobgen-psql/driver/parser/verify.go
atzedus: Some checks added

stephenafamo:
A few CHANGELOG conflicts to resolve. @atzedus

atzedus:
Resolved

PR #663, opened 2026-05-01
fix(factory): prevent stack overflow from circular references in FromExisting methods

### LINKED ISSUE #584: Stack Overflow in Factory `FromExisting*` Methods Due to Circular References
## Description

The generated factory code's `FromExisting*` methods cause infinite recursion and stack overflow when using models that were created with bidirectional relationships populated in their `.R` struct.

## Environment

- Bob version: v0.41.1
- Database: PostgreSQL
- Go version: 1.23+

## Database Schema

```sql
CREATE TABLE orders (
    id SERIAL PRIMARY KEY,
    name TEXT NOT NULL
);

CREATE TABLE items (
    id SERIAL PRIMARY KEY,
    order_id INT NULL,
    description TEXT NOT NULL,
    FOREIGN KEY (order_id) REFERENCES orders(id)
);

CREATE TABLE item_details (
    id SERIAL PRIMARY KEY,
    item_id INT NOT NULL,
    details TEXT NOT NULL,
    FOREIGN KEY (item_id) REFERENCES items(id)
);
```

## Problem

When an `Order` is created with items using the factory's relationship helpers, the bidirectional relationships are automatically populated:
- `orderDao.R.Items[0]` contains the item
- `orderDao.R.Items[0].R.Order` points back to `orderDao`

When attempting to use this item with a `WithExisting*` modifier:

```go
// Create order with items
orderDao := factory.NewOrder(
    dbfactory.OrderMods.AddNewItems(2),
).CreateOrFail(ctx, t, db)

// This causes stack overflow because orderDao.R.Items[0].R.Order points back to orderDao
factory.NewItemDetail(
    dbfactory.ItemDetailMods.WithExistingItem(orderDao.R.Items[0])
).CreateOrFail(ctx, t, db)
```

This results in:

```
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0x14031200380 stack=[0x14031200000, 0x14051200000]
fatal error: stack overflow
```

## Root Cause

The generated `FromExisting*` methods recursively load ALL relationships without checking for circular references:

```go
func (f *Factory) FromExistingItem(m *models.Item) *ItemTemplate {
    o := &ItemTemplate{f: f, alreadyPersisted: true}
    
    // ... field assignments ...
    
    ctx := context.Background()
    if m.R.Order != nil {
        ItemMods.WithExistingOrder(m.R.Order).Apply(ctx, o)  // ← Calls FromExistingOrder
    }
}

func (f *Factory) FromExistingOrder(m *models.Order) *OrderTemplate {
    o := &OrderTemplate{f: f, alreadyPersisted: true}
    
    // ... field assignments ...
    
    ctx := context.Background()
    if len(m.R.Items) > 0 {
        OrderMods.AddExistingItems(m.R.Items...).Apply(ctx, o)  // ← Calls FromExistingItem for each
    }
}
```

This creates an infinite loop: **Item → Order → Items → Order → ∞**

The issue occurs because:
1. `orderDao.R.Items[0]` has its `R.Order` field populated (pointing back to `orderDao`)
2. `WithExistingItem()` internally calls `FromExistingItem()`
3. `FromExistingItem()` sees `m.R.Order != nil` and calls `WithExistingOrder(m.R.Order)`
4. `WithExistingOrder()` calls `FromExistingOrder()` 
5. `FromExistingOrder()` sees `m.R.Items` and calls `AddExistingItems()` for each item
6. This loops back to step 2, creating infinite recursion

## Current Workaround

Manually break the circular reference before using the model:

```go
orderDao.R.Items[0].R.Order = nil
factory.NewItemDetail(
    dbfactory.ItemDetailMods.WithExistingItem(orderDao.R.Items[0])
).CreateOrFail(ctx, t, db)
```

Or re-query the model without preloading relationships:

```go
itemsDao, err := dbmodels.Items.Query(
    dbmodels.SelectWhere.Items.OrderID.EQ(orderDao.ID),
    // Don't preload relationships - this prevents R.Order from being populated
).All(ctx, db)
require.NoError(t, err)

factory.NewItemDetail(
    dbfactory.ItemDetailMods.WithExistingItem(itemsDao[0])
).CreateOrFail(ctx, t, db)
```

## Additional Context

This issue affects any schema with bidirectional relationships:
- Parent-Child relationships
- Many-to-many relationships with join tables
- Any circular relationship patterns

The issue is particularly problematic in test code where factories are heavily used with pre-existing test data that has relationships already loaded in the `.R` struct.

wucm667:
Fixes #584

## Problem

The generated factory code's `FromExisting*` methods cause infinite recursion and stack overflow when using models with bidirectional relationships populated in their `.R` struct.

When `orderDao.R.Items[0].R.Order` points back to `orderDao`, calling `FromExistingItem` triggers:
```
Item -> Order -> Items -> Order -> Items -> ... (infinite loop)
```

## Solution

Added a `visited sync.Map` field to the `Factory` struct to track model pointers during `FromExisting*` calls. Before processing relationships, the method checks if the model pointer has already been visited. If so, it returns early with field values only (no relationship processing), breaking the circular reference chain.

## Changes

- `gen/templates/factory/bobfactory_main.bob.go.tpl`: 
  - Added `visited sync.Map` field to `Factory` struct
  - Added `sync` and `unsafe` imports
  - In `FromExisting*` methods, check `visited` before processing relationships
  - Use `defer f.visited.Delete(ptr)` to clean up after processing

## Example

Before (stack overflow):
```go
orderDao := factory.NewOrder(factory.OrderMods.AddNewItems(2)).CreateOrFail(ctx, t, db)
// This caused stack overflow because orderDao.R.Items[0].R.Order points back to orderDao
factory.NewItemDetail(dbfactory.ItemDetailMods.WithExistingItem(orderDao.R.Items[0])).CreateOrFail(ctx, t, db)
```

After (works correctly):
```go
// Same code now works - circular reference is detected and skipped
```

### DIFF
--- a/gen/templates/factory/bobfactory_main.bob.go.tpl
+++ b/gen/templates/factory/bobfactory_main.bob.go.tpl
@@ -1,7 +1,11 @@
 {{$.Importer.Import "context"}}
 {{$.Importer.Import "models" (index $.OutputPackages "models") }}
+{{$.Importer.Import "sync"}}
+{{range $table := .Tables}}{{if $.Relationships.Get $table.Key}}{{$.Importer.Import "unsafe"}}{{end}}{{end}}
 
 type Factory struct {
+    // visited tracks model pointers during FromExisting calls to prevent circular reference stack overflow
+    visited sync.Map
     {{range $table := .Tables}}
     {{ $tAlias := $.Aliases.Table $table.Key -}}
 		base{{$tAlias.UpSingular}}Mods {{$tAlias.UpSingular}}ModSlice
@@ -40,17 +44,23 @@ func (f *Factory) FromExisting{{$tAlias.UpSingular}}(m *models.{{$tAlias.UpSingu
   {{end}}
 
   {{if $.Relationships.Get $table.Key -}}
-  ctx := context.Background()
+  // Check for circular references using Factory-level visited map to prevent stack overflow
+  // See https://github.com/stephenafamo/bob/issues/584
+  ptr := uintptr(unsafe.Pointer(m))
+  if _, loaded := f.visited.LoadOrStore(ptr, struct{}{}); loaded {
+    return o // Already processing this model, skip to prevent infinite recursion
+  }
+  defer f.visited.Delete(ptr) // Clean up after processing
   {{- end}}
   {{range $.Relationships.Get $table.Key -}}
     {{$relAlias := $tAlias.Relationship .Name -}}
     {{if .IsToMany -}}
       if len(m.R.{{$relAlias}}) > 0 {
-      {{$tAlias.UpSingular}}Mods.AddExisting{{$relAlias}}(m.R.{{$relAlias}}...).Apply(ctx, o)
+      {{$tAlias.UpSingular}}Mods.AddExisting{{$relAlias}}(m.R.{{$relAlias}}...).Apply(context.Background(), o)
       }
     {{- else -}}
       if m.R.{{$relAlias}} != nil {
-      {{$tAlias.UpSingular}}Mods.WithExisting{{$relAlias}}(m.R.{{$relAlias}}).Apply(ctx, o)
+      {{$tAlias.UpSingular}}Mods.WithExisting{{$relAlias}}(m.R.{{$relAlias}}).Apply(context.Background(), o)
       }
     {{- end}}
   {{end}}
--- a/gen/templates/factory/bobfactory_main.bob.go.tpl
+++ b/gen/templates/factory/bobfactory_main.bob.go.tpl
@@ -13,7 +13,6 @@ func New() *Factory {
   return &Factory{}
 }
 
-type visitedCtxKey struct{}
 
 {{range $table := .Tables}}
 {{ $tAlias := $.Aliases.Table $table.Key -}}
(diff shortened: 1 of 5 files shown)


stephenafamo:
By having the `sync.Map` on the Factory itself, wouldn't that possibly cause conflicts if used concurrently?

wucm667:
Good catch. I moved the visited map from the Factory level to a per-call context-based approach. Each top-level FromExistingXxx call creates its own visited map and passes it through context to recursive calls (including WithExistingXxx and AddExistingXxx). This means goroutines are fully isolated and a regular map is sufficient -- no sync.Map needed since each map stays within a single call chain.

gen/templates/factory/bobfactory_main.bob.go.tpl
```diff
@@ -12,6 +13,8 @@ func New() *Factory {
   return &Factory{}
 }
 
+type visitedCtxKey struct{}
```
stephenafamo: Perhaps this should be moved to `bobfactory_context.bob.go.tpl` and match the context pattern there

gen/templates/factory/bobfactory_main.bob.go.tpl
```diff
@@ -31,6 +34,16 @@ func (f *Factory) New{{$tAlias.UpSingular}}WithContext(ctx context.Context, mods
 }
 
 func (f *Factory) FromExisting{{$tAlias.UpSingular}}(m *models.{{$tAlias.UpSingular}}) *{{$tAlias.UpSingular}}Template {
+  {{if $.Relationships.Get $table.Key -}}
+  visited := make(map[uintptr]struct{})
+  ctx := context.WithValue(context.Background(), visitedCtxKey{}, visited)
+  {{- else -}}
+  ctx := context.Background()
```
stephenafamo: I'm am generally against creating fresh context instances within library code.

If there is no other option, I would rather change `FromExisting` to take a context.

stephenafamo (changes requested):

stephenafamo:
Also remember to update CHANGELOG.md and include attribution for the changes.

wucm667:
@stephenafamo Added the changelog entry as requested.

stephenafamo:
@wucm667 see my other comments

wucm667:
@stephenafamo Done — adopted both suggestions:
- Moved the context key to  using the  pattern
-  now accepts  as the first parameter, no more internal

stephenafamo (approved):

PR #665, opened 2026-05-03
fix: a relation should only be required if all columns require

jay-babu:
## What changed

This updates generated relation metadata so a relationship is marked required only when all columns in the relationship are required. Relationships without foreign-key columns are explicitly treated as not required.

It keeps `WithParentsCascading()` cascading through all to-one relationships when explicitly requested.

## Why

A relation with mixed required and optional columns should not be treated as required as soon as any single column is required. However, explicit parent cascading should still include optional parents because generated factory and unique-constraint workflows rely on that behavior.

## Validation

- `go test ./gen/drivers ./gen`
- `go test ./gen/bobgen-sqlite/driver -run 'TestSQLite/driver/modernc/generate' -count=1 -timeout=20m`
- `go test ./...` was also run; it hit Go's default 10m timeout in `TestSQLite/driver/modernc/generate`, and the timed-out subtest passed when rerun with a longer timeout.


stephenafamo:
I don't think this is correct behaviour at all.

rajp33:
@stephenafamo why do you think so?

stephenafamo:
Because as far as I know, even if one column of a foreign key is null, the other columns MUST be provided

PR #666, opened 2026-05-03
fix: template for random_factory imports unnecessarily

jay-babu:
## What changed

This narrows the random factory template import condition so it only imports when the generated code actually needs the dependency.

## Why

The previous template condition could include an unnecessary import in generated random factory code.

## Validation

- `go test ./gen`


stephenafamo:
Add a note to CHANGELOG.md

jay-babu:
@stephenafamo done

PR #670, opened 2026-05-06
feat: add Unqualified() and UnquotedLast() methods to Quote expressions

atzedus:
- Added `Unqualified()` method to quoted expressions returned by `Quote()` — returns a new expression containing only the last identifier part
- Added `UnquotedLast()` method to extract the last identifier as a string
- Both methods available on all dialect quote expressions (psql, mysql, sqlite)

**Why this is needed**

Generated column structures use `Quote()` extensively to create qualified column references like `"schema"."table"."column"`. The new methods allow users to extract just the unqualified column name without the full qualification when needed.

**Example usage**

```go
// Old approach
mm.Columns("status") // not type-safe or dbinfo plugin needed to use

// New approach with extracted unqualified name
mm.Columns(dbmodels.AuditEvents.Columns.Status.UnquotedLast())

sm.Where(dbmodels.Users.Columns.ID.Unqualified().EQ(psql.Arg(1)) // SQL: "id" = $1
```

This avoids accidental double-quoting and makes the generated column API more ergonomic. The `Unqualified()` method returns an expression that can be further chained or rendered as SQL, while `UnquotedLast()` provides direct string access for simple cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added PostgreSQL `MERGE` statement SQL parser used for sql-to-code generation (thanks @atzedus)
+- Added `Unqualified()` method to quoted expressions returned by `Quote()` across all dialects. Returns a new expression containing only the last identifier part. (thanks @atzedus)
+- Added `UnquotedLast()` method to quoted expressions for extracting the last identifier as a string. (thanks @atzedus)
 
 ### Changed
 
--- a/expr/chain.go
+++ b/expr/chain.go
@@ -16,6 +16,23 @@ func (x Chain[T, B]) WriteSQL(ctx context.Context, w io.StringWriter, d bob.Dial
 	return bob.Express(ctx, w, d, start, x.Base)
 }
 
+// Unqualified returns a new expression containing only the last qualified part.
+// If the base expression doesn't support Unqualified, returns the base expression unchanged.
+func (x Chain[T, B]) Unqualified() T {
+	if u, ok := x.Base.(interface{ Unqualified() bob.Expression }); ok {
+		return X[T, B](u.Unqualified())
+	}
+	return X[T, B](x.Base)
+}
+
+// UnquotedLast returns the last element as a raw unquoted string, or empty string if not supported.
+func (x Chain[T, B]) UnquotedLast() string {
+	if u, ok := x.Base.(interface{ UnquotedLast() string }); ok {
+		return u.UnquotedLast()
+	}
+	return ""
+}
+
 // IS DISTINCT FROM
 func (x Chain[T, B]) IsDistinctFrom(exp bob.Expression) T {
 	return X[T, B](Join{Exprs: []bob.Expression{x.Base, isDistinctFrom, exp}})
--- a/expr/quote.go
+++ b/expr/quote.go
@@ -13,6 +13,7 @@ func Quote(aa ...string) bob.Expression {
 		if v == "" {
 			continue
 		}
+
 		ss = append(ss, v)
 	}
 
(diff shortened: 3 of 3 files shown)


stephenafamo:
Honestly, I'm not sure about this, especially since this can be done with the `dbinfo` plugin.

Both methods seems specific to "quote" and then to have it on `Chain` feels somewhat wrong.

atzedus:
Thanks for the feedback! I agree that having methods on the Chain type might feel a bit unconventional at first glance. The main reason for this design is ergonomics: without methods on Chain, we end up with code like this:

```go
q := psql.Quote("schema", "table", "id")
// q is psql.Expression, but Unqualified() is not available
// Would need type assertion: if qu, ok := q.Base.(quoted) { ... }
```

With methods on Chain, it becomes much cleaner:

```go
q := psql.Quote("schema", "table", "id")
unq := q.Unqualified()  // Works directly!
```

Regarding the dbinfo plugin — I totally see your point, and in theory it covers this use case. However, in practice (especially when working with fully generated code for INSERT/MERGE statements), ~95% of query building ends up using models and helpers from the models plugin. The dbinfo plugin is only needed in rare cases just to get type-safe access to methods like im.SetCol(from string), mm.Columns(columns ...string), or im.Into(name any, columns ...string).
This leads to maintaining two parallel sets of variables:

```go
usersCols := dbmodels.Users.Columns      // from models plugin
usersColsRaw := dbinfo.Users.Columns     // from dbinfo plugin
```

And when using dbinfo, you always need to access the .Name field explicitly (e.g., usersColsRaw.Login.Name), which adds visual noise and feels a bit verbose when dealing with many fields.
That said, I'm open to alternatives if there's a way to keep the API ergonomic without extending Chain. Happy to discuss further!

atzedus:
> Both methods seems specific to "quote" and then to have it on `Chain` feels somewhat wrong.

We can move `Unqualified()` and `UnquotedLast()` out of `expr.Chain` and define them on each `dialect.Expression` instead. That keeps Chain generic and removes quote-specific API from the shared base.

That said, the practical difference is mostly organizational: each `dialect.Expression` still embeds `expr.Chain`, so behavior is still chain-based under the hood.

What do you think about adding these methods at all?
Do you see this as useful API surface in bob?

@stephenafamo

stephenafamo:
Overall, I see the point, I see how it could be useful.  
On the other hand, I am a bit wary of adding more and more methods, I fear that Bob could become quite bloated.

Which is why I would prefer at most one way to do something.

It is also part of the reason why I adopted the plugin architecture for code generation so that users can pick and choose what they want.


Summary: I see the usefulness of this, my main hesitation is regarding bloat.

atzedus:
I think you're right about the risk of bloating the core. Let me take a step back and approach this from the `models` plugin side instead.

Perhaps it would be cleaner to improve the structure and the method that generates column helpers, something like this:

```go
func buildUserColumns(tableName string) userColumns {
	columnsExpr := expr.NewColumnsExpr(
		"id", "login", "email",
	)
	if tableName != "" {
		columnsExpr = columnsExpr.WithParent(tableName)
	}
	return userColumns{
		ColumnsExpr: columnsExpr,
		tableAlias:  tableName,
		ID:          psql.Quote(tableName, "id"),
		Login:       psql.Quote(tableName, "login"),
		Email:       psql.Quote(tableName, "email"),
	}
}
```

We already added the `Unqualified()` method earlier. I think methods like `Unquoted()` (or similar), scoped specifically to a column within this generated structure and accessible only from here, would be more appropriate in this context.

One of the main drivers for this is the `models.Users.Columns.Except(cols ...string)` method. To use it with type safety, we currently have to reach into the `dbinfo` package just to get the string representation of a column, even though all the necessary information is already available in `models`.

If you don't have objections to this direction, I'll try to prepare a separate PR focused on the `models` plugin later.

@stephenafamo

atzedus:
@stephenafamo
What do you think about this: https://github.com/stephenafamo/bob/pull/675

The generated structure for Columns is already a wrapper over expr.ColumnsExpr, so I believe the individual fields themselves have every right to be wrappers as well.

This pull request is no longer relevant - the new one fully addresses all the outstanding points. I think we can go ahead and close the current PR.

stephenafamo:
I like the other PR. Thank you for giving it more thought.

PR #675, opened 2026-05-07
feat: add unqualified column type API and migration notes

atzedus:
This PR introduces a new generated column type API for columns.

**Changes**
- Generated column accessors now return a dialect-specific column type that embeds `dialect.Expression`.
- Added helper methods on generated columns:
  - `Alias()`
  - `Name()`
  - `AliasedAs(alias)`
  - `Unqualifed()`

**Breaking change**
Previously, generated column fields returned `dialect.Expression` directly.  
Now they return a column wrapper type that embeds `dialect.Expression`.

Migration is simple: when an API expects an expression, use `.Expression`.

```go
// before
where.EQ(models.Users.Columns.ID, 1)

// after
where.EQ(models.Users.Columns.ID.Expression, 1)
```

If you have a function that expects (for example) `psql.Expression`:

```go
func Contains(field psql.Expression, s string) bob.Expression {
	return field.ILike(psql.Arg(s))
}
```

Update it to accept the more general `bob.Expression` interface:

```go
func Contains(field bob.Expression, str string) bob.Expression {
	return dialect.NewExpression(field).ILike(psql.Arg(s))
}
```

Now it works with both the new column types and any other expression:

```go
// Works with new column type API:
Contains(models.Users.Columns.Email, "%example%")

// Works with legacy expressions:
Contains(models.Users.Columns.Email.Expression, "%example%")
```

**Why `Name()` is useful**
`Name()` is convenient when you need raw column names, for example with `models.Users.Columns.Except(...)` to build dynamic column sets.

```go
excluded := models.Users.Columns.Except(
    models.Users.Columns.ID.Name(),
    models.Users.Columns.CreatedAt.Name(),
)
```

`Alias()`, `AliasedAs(alias)`, `Unqualifed()` the same as for *Columns type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added PostgreSQL `MERGE` statement SQL parser used for sql-to-code generation (thanks @atzedus)
 - Added `As(alias)` method to `bob.BaseQuery`, allowing queries (e.g. `mysql.Select(...)`) to be aliased directly when used as subqueries in a column list. (thanks @jacobmolby)
 - Added unqualified columns API: generated column structures now provide Alias(), Name(), and AliasedAs(alias) methods for easier column metadata access and aliasing. (thanks @atzedus)
+- Added `bob.ParensOmitter` interface with `ShouldOmitParens() bool` to let expressions opt out of automatic parenthesis wrapping in expression builders. (thanks @atzedus)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added PostgreSQL `MERGE` statement SQL parser used for sql-to-code generation (thanks @atzedus)
 - Added `As(alias)` method to `bob.BaseQuery`, allowing queries (e.g. `mysql.Select(...)`) to be aliased directly when used as subqueries in a column list. (thanks @jacobmolby)
-- Added unqualified columns API: generated column structures now provide Alias(), Name(), and AliasedAs(alias) methods for easier column metadata access and aliasing. (thanks @atzedus)
+- Added unqualified columns API: generated column structures now provide Alias(), Name(), AliasedAs(alias) and Unqualified() methods for easier column metadata access and aliasing. (thanks @atzedus)
 - Added `bob.ParensOmitter` interface with `ShouldOmitParens() bool` to let expressions opt out of automatic parenthesis wrapping in expression builders. (thanks @atzedus)
 
 ### Changed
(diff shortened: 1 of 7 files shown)


stephenafamo:
I like this direction much more.

I think there is a way to do this without making it a breaking change. 

* Directly implement the `Expression` interface and call the `Expression`'s `WriteSQL` method.
* Add a `DoNotWrap` interface to `Bob`, implement it in this type, and add it as a special case in `expr/builder.go` to prevent adding unnecessary parenthesis

I believe this should avoid any breaking changes right?

CHANGELOG.md
```diff
@@ -11,12 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - **BREAKING:** `mm.SetCol()` now returns `mods.Set[*mm.UpdateAction]` instead of a custom `SetChain` type. `.ToExpr(val)` is replaced by `.To(val)`, and `.ToDefault()` is replaced by `.To(psql.Raw("DEFAULT"))`. `.To()` and `.ToArg()` work as before. (thanks @atzedus)
 - **BREAKING:** `mm.Recursive()` has been removed. PostgreSQL does not support `WITH RECURSIVE` in MERGE statements. (thanks @atzedus)
+- **BREAKING:** Generated column accessor fields in *Columns structs (e.g., models.Users.Columns.ID, models.Users.Columns.Name) now return a dialect-specific column type (e.g., userColumn) that embeds dialect.Expression, instead of directly returning dialect.Expression. This allows access to the new Alias(), Name(), and AliasedAs() methods. For code that previously used the column accessors directly as expressions, simply add .Expression to access the underlying expression: models.Users.Columns.ID becomes models.Users.Columns.ID.Expression. (thanks @atzedus)
```
stephenafamo: CHANGELOG is now outdated

dialect/psql/merge_test.go
```diff
@@ -9,6 +9,12 @@ import (
 	testutils "github.com/stephenafamo/bob/test/utils"
 )
 
+type wrappedExpr struct {
+	psql.Expression
+}
+
+func (wrappedExpr) ShouldOmitParens() bool { return true }
+
```
stephenafamo: This should probably be tested in `expressions_test.go`

atzedus:
> * Directly implement the `Expression` interface and call the `Expression`'s `WriteSQL` method.

I don't think this is strictly necessary - the new type embeds `dialect.Expression` directly, so it automatically satisfies the `bob.Expression` interface via composition. The WriteSQL method should be delegated correctly without additional boilerplate.

> * Add a `DoNotWrap` interface to `Bob`, implement it in this type, and add it as a special case in `expr/builder.go` to prevent adding unnecessary parenthesis

Great catch on the parentheses issue - that's exactly the kind of edge case I wanted to handle cleanly. I've implemented this idea using a `bob.ParensOmitter` interface, with a `ShouldOmitParens()` bool method to give the builder fine-grained control. You can see the changes here:
https://github.com/stephenafamo/bob/pull/675/changes/64f1b0531793c30d24e9f055cafd90d3e984f2b5#diff-890293cf39fef9ca0546a39df6c563073624cf349f8ec3c27aa5ee24dd72af3d
Regarding breaking changes: I agree we should avoid them where possible, but in this case they're limited to a very narrow edge case. The only code affected would be external consumers that explicitly type-assert or depend on the concrete `psql.Expression` type (which isn't an interface). Since most well-structured code should depend on the `bob.Expression` interface instead, I believe this impact is both easy to fix and extremely rare in practice.

Happy to discuss further if you see other ways to soften the transition!

dialect/psql/merge_test.go
atzedus: done

CHANGELOG.md
atzedus: I've updated the description to reflect the changes.

stephenafamo:
Great work. 

I really appreciate the effort you've put into this. Thank you

PR #679, opened 2026-05-13
Remove Unqualified() method from generated *Columns structs

jacobmolby:
Since #675 was merged, the changes made in #625 are no longer needed.

I think it's better to remove it sooner, rather than later so two almost identical APIs don't need to be supported for ever.

This PR basically reverts #625.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added PostgreSQL `RETURNING WITH (OLD AS ..., NEW AS ...)` support for `INSERT`, `UPDATE`, `DELETE`, and `MERGE` query builders, including fluent modifiers: `im/um/dm/mm.Returning(...).WithOldAs(...).WithNewAs(...)`. (thanks @atzedus)
   - Note: `bobgen-psql` sql-to-code parsing for this syntax is currently blocked by upstream PostgreSQL parser dependency support.
 
+### Removed
+
+- Removed the `Unqualified()` method from generated `*Columns` (plural) structs. It is superseded by the per-column API introduced in the unqualified columns API entry above: use `Column.Unqualified()` for a single column to build an unqualified reference. (thanks @jacobmolby)
+
 ### Changed
 
 - **BREAKING:** Renamed the generated `ThenLoadCount` variable to `SelectThenLoadCount` for naming consistency with `SelectThenLoad`/`InsertThenLoad`/`UpdateThenLoad`. Update call sites from `models.ThenLoadCount.X.Y` to `models.SelectThenLoadCount.X.Y`. (thanks @jacobmolby)
--- a/gen/templates/models/table/001_types.go.tpl
+++ b/gen/templates/models/table/001_types.go.tpl
@@ -78,7 +78,7 @@ func build{{$tAlias.UpSingular}}Columns(alias string) {{$tAlias.DownSingular}}Co
   return {{$tAlias.DownSingular}}Columns{
     ColumnsExpr: expr.NewColumnsExpr(
       {{range $column := $table.Columns -}}{{quote $column.Name}},{{end}}
-    ).WithParent({{quote $table.Key}}),
+    ).WithParent(alias),
     tableAlias: alias,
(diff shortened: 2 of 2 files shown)


atzedus:
I am convinced that this functionality should be retained. First of all, removing it would be a breaking change, especially since it is not fully replaced by the new implementation (introduced in https://github.com/stephenafamo/bob/pull/675).

For the `*Columns` type, `Unqualified()` is a more intuitive alternative to the `AliasedAs("")` construct.

Here's an example where `Unqualified()` on `*Columns` is necessary:

```go
cols := models.Users.Columns

psql.Insert(
	im.Into(models.Clients.Name()),
	im.OnConflict(cols.Unqualified().DisableAlias().Except("id")),
)
// generates something like: INSERT INTO "clients" DEFAULT VALUES ON CONFLICT ("name", "inn", "kpp")
```

The same example, but for a single column using `Unqualified()` on `*Column`:

```go
cols := models.Users.Columns

psql.Insert(
	im.Into(models.Clients.Name()),
	im.OnConflict(cols.ID.Unqualified()),
)
// generates something like: INSERT INTO "clients" DEFAULT VALUES ON CONFLICT ("id")
```

Generally speaking, `cols.Unqualified().ID` and `cols.ID.Unqualified()` are equivalent. In that light, I would lean toward removing `Unqualified()` from the `*Column` type. However, I still think it could be useful in other, less common scenarios when working with a single column.

stephenafamo:
I think it is best to keep it for now since like @atzedus said, they new `Unqualified()` method on individual columns cannot fully replace it.



> Generally speaking, `cols.Unqualified().ID` and `cols.ID.Unqualified()` are equivalent. In that light, I would lean toward removing `Unqualified()` from the `*Column` type.

I agree with this, and I feel the same.  
Can you think of the situations where having the `Unqualified()` method on the `*Column` type cannot be replaced by the the top level `Unqualified()` method? @atzedus

atzedus:
> I agree with this, and I feel the same. Can you think of the situations where having the `Unqualified()` method on the `*Column` type cannot be replaced by the the top level `Unqualified()` method?

At the moment, I don't see a use case where this would be required. The main reason to keep `Unqualified()` on `*Column` is performance: calling it on `*Column` has significantly less overhead than calling it on `*Columns`, since the latter rebuilds the entire collection just to work with a single column.

stephenafamo:
Okay, let's remove it then, which I believe means reverting #675

atzedus:
> Okay, let's remove it then, which I believe means reverting #675

No, no - I only meant the single `Unqualified()` method from #675. All the rest of its functionality is highly useful and necessary, and there's no good replacement for it.

If the need for the `Unqualified` method from #675 arises later, it can be easily added since it follows the same pattern as for `*Columns`. It's essentially just a clearer alternative to calling `AliasedAs("")`.

The most critical method added in #675 is `Name()`; everything else was included solely for consistency with the existing `*Columns` API.

stephenafamo:
Alright, then just a small PR to remove the `Unqualified` method.

stephenafamo:
I'll close this PR in the mean time

jacobmolby:
Makes sense to keep it. I didn't think of the case for multiple cols being used at once.

atzedus:
Currently, the equivalent calls are as follows:

```go
cols := models.Users.Columns

cols.Unqualified().ID = cols.ID.Unqualified()
cols.AliasedAs("test").ID = cols.ID.AliasedAs("test")
cols.AliasedAs("test").Alias() = cols.ID.AliasedAs("test").Alias()
```

Therefore, these are exactly the ones that can be removed in another small PR.

```go
cols := models.Users.Columns
cols.ID.Name() // no alternative replacement available
```

However, if we remove `Unqualified()` and `AliasedAs()` from `*Column`, then the call

```go
cols.AliasedAs("test").ID.Alias()
```

also seems to lose its purpose, since we would lose the ability to change the `Alias` for a single specific column within the entire set — we could only change it for all columns at once.

As I mentioned earlier, the most essential and core functionality is the `Name()` method for a single specific column; everything else can be removed for now, since the need for it is currently unclear to anyone.

I'll create a PR a bit later to clean up the unnecessary parts.

PR #685, opened 2026-05-14
feat(psql): add grouping-element helpers for GROUP BY

atzedus:
This PR adds PostgreSQL grouping-element helpers for GROUP BY: grouped tuple, ROLLUP, CUBE, and GROUPING SETS.
It extends clause primitives and psql helpers, with focused select tests for each grouping form.


stephenafamo:
Looks good.

PR #686, opened 2026-05-14
feat(expr): Add ColumnsExpr accessors and improve column rendering defaults

atzedus:
## Updates

**Note:** An earlier iteration of this PR introduced `ColumnsExpr.Expressions()` method and a dedicated `expr.Expressions` type. These have been rolled back in favor of a simpler API focused on accessor methods and smarter default rendering behavior. The core benefit—proper handling of column qualification with cleaner defaults—remains fully intact.

**Below is the description of the original PR:**

This PR adds ColumnsExpr.Expressions() and updates related behavior around column rendering and code generation.

What is included:

- ColumnsExpr.Expressions() returns expr.Expressions (a dedicated expression list type).
- expr.Expressions preserves current ColumnsExpr options for each generated item:
  - parent qualification
  - aggFunc wrappers
  - aliasPrefix
  - aliasDisabled
- Default rendering is now cleaner:
  - when parent is effectively empty, aliasPrefix is empty, and aggFunc is empty, output is plain quoted column name (for example "name"), not redundant self-aliasing ("name" AS "name").
- The template-side DisableAlias workaround is no longer needed.
  - This effectively rolls back the related change introduced in https://github.com/stephenafamo/bob/pull/683.

## Where this is useful

Especially useful with sm.Columns(clauses ...any) and generated model columns:

```go
columns := models.Users.Columns.Except("id").Expressions()
sm.Columns(columns.Any()...)
```

## Compatibility note

expr.Expressions is fully compatible with bob.Expression.
You can pass it directly or convert explicitly with a regular type conversion:

```go
columns := models.Users.Columns.Except("id").Expressions()

var asExpr bob.Expression = columns
// explicit conversion form:
asExpr = bob.Expression(columns)

sm.Columns(asExpr)
```

## Important distinction

This is not equivalent to using Names():

```go
columns := toAnySlice(models.Users.Columns.Except("id").Names())
sm.Columns(columns...)
```

Names() returns raw/unqualified string names, while Expressions() returns SQL-ready quoted expressions with current ColumnsExpr behavior applied.


stephenafamo:
I don't think this is useful since the `ColumnsExpr` type is a `bob.Expression` itself that expands to the list without first turning it into a slice.

> Especially useful with sm.Columns(clauses ...any) and generated model columns:
> 
> ```go
> columns := models.Users.Columns.Except("id").Expressions()
> sm.Columns(columns.Any()...)
> ```

It should be possible to already use the existing `ColumnsExpr` like this:

```go
columns := models.Users.Columns.Except("id")
sm.Columns(columns)
```

What is perhaps useful is to disable alias if there is no parent.

atzedus:
@stephenafamo Thanks for the quick feedback. You are absolutely right — `ColumnsExpr` is already a `bob.Expression` and can be passed directly to `sm.Columns(columns)`. Let me address your points and clarify my motivation:

1. **Regarding `Expressions()` and `sm.Columns(...any)`**:
   Yes, it works as you described. The only motivation behind adding `Expressions()` was semantic clarity: `sm.Columns(clause ...any)` implies that each element of the slice is a *single* column expression, whereas `ColumnsExpr` represents a *set* of columns (comma-separated). Syntactically, the resulting SQL looks correct either way, but if `sm.Columns` ever assumes that each argument is a single-column expression (e.g., for validation or introspection), passing a "set" disguised as multiple args could theoretically cause issues down the line. That said, if this is over-engineering, I'm happy to drop the `Expressions()` addition entirely.

2. **Regarding `("name" AS "name")`**:
   I consider this specific case (redundant self-aliasing when there's no parent/table prefix) absolutely useless and noisy in the final SQL. I'd like to keep the part of this PR that automatically suppresses aliases in that scenario — it makes the output cleaner without changing semantics.

3. **Regarding `DisableAlias()` method**:
   Given the current changes and the fact that `disableAlias = false` by default (with automatic suppression of redundant aliases), I believe the `DisableAlias()` method is now obsolete. I searched the codebase and didn't find a single usage of it. Unless there's a hidden use case I'm missing, I propose removing it altogether to reduce API surface.

4. **Regarding `WithAgg()` method**:
   I'd really appreciate some insight here — in what scenario is `WithAgg()` useful? I couldn't find any usage examples or unit tests for it in the codebase. If it's legacy or unused, maybe it's a candidate for removal as well?

**To summarize, how would you like to proceed with this PR?**
1. Drop the `Expressions()` method and related code?
2. Keep the automatic suppression of redundant aliases like `("name" AS "name")`?
3. Remove the `DisableAlias()` method?
4. Remove the `WithAgg()` method and related code?

Happy to adjust the PR based on your guidance.

stephenafamo:
> 1. **Regarding `Expressions()` and `sm.Columns(...any)`**:

Yes, I would prefer if it is dropped. For things like validation and introspection, it is always best to assume that an expression can expand to "anything" and only the full expansion can be authoritative.

> 2\. **Regarding `("name" AS "name")`**:

Yes, let's keep this part of the PR.

> 3\. **Regarding `DisableAlias()` method**:

Redundant aliases my be auto-supressed, but there may be cases where the user wants `t.id, t.name, t.age`.  
I'm am much more hesitant about removing this one


> 4\. **Regarding `WithAgg()` method**:

This is very useful (perhaps it needs a test). There can be situations where the user wants to wrap each column in a function.  
Without the `WithAgg` method, this becomes essentially impossible, except the string names are retrieved and individually wrapped.

atzedus:
> > 3. **Regarding `DisableAlias()` method**:
> 
> Redundant aliases my be auto-supressed, but there may be cases where the user wants `t.id, t.name, t.age`. I'm am much more hesitant about removing this one

I completely agree, I didn't think about this case.

> > 4. **Regarding `WithAgg()` method**:
> 
> This is very useful (perhaps it needs a test). There can be situations where the user wants to wrap each column in a function. Without the `WithAgg` method, this becomes essentially impossible, except the string names are retrieved and individually wrapped.

I've been thinking that instead of this, a hooks mechanism would probably be more useful. For example, it could allow the `sum` field with `parent` = `t` to be transformed into `(t.sum * t.sum + 100) AS "sum"` on output. But I think that's a topic for a separate PR. Alright, let's leave it as is for now.

One last question I wanted to clarify:

`ColumnsExpr` currently has a `Names()` method that essentially just returns the internal `names` field, so it acts as a simple getter. I'd like to add similar getters for the other fields in this PR: `Parent()`, `AggFunc()`, `AliasPrefix()`, and `AliasDisabled()`.

Do you think this is a good idea? I see the value in having them, as they would make it straightforward to implement `Expressions` as a wrapper around `ColumnsExpr` if needed in the future.

@stephenafamo

atzedus:
> > > 3. **Regarding `DisableAlias()` method**:
> > 
> > 
> > Redundant aliases my be auto-supressed, but there may be cases where the user wants `t.id, t.name, t.age`. I'm am much more hesitant about removing this one
> 
> I completely agree, I didn't think about this case.

I got that backwards—I actually meant `EnableAlias()`. By default, it already behaves as if it were called, so you typically only need `DisableAlias()` when necessary. Right now, I only see a practical use for `EnableAlias()` in something like `ColumnsExpr.DisableAlias().EnableAlias()` 😆 But fair enough, let's leave it as is for now.

@stephenafamo

atzedus:
@stephenafamo 

I've pushed all the discussed changes to this PR, assuming they look good to you. It's now ready to be merged. If anything else needs tweaking, I'm happy to discuss or make further adjustments.

stephenafamo:
Looks good. Kindly update the PR title 🙏🏾

atzedus:
> Looks good. Kindly update the PR title 🙏🏾

Done, also description updated.

PR #688, opened 2026-05-18
feat(psql, sqlite): add im.Excluded helper and NoSep constant for expr.Join

atzedus:
This pull request introduces the `Excluded` helper for PostgreSQL and SQLite. These changes ensure correct handling of `EXCLUDED.<col>` expressions and improve code reusability.

#### Problem
Unwanted spaces in concatenated SQL expressions caused incorrect rendering, particularly for `EXCLUDED.<col>` constructs.

#### Solution
1. **New Helper**: `Excluded` simplifies generating `EXCLUDED.<col>` expressions.
   - Example:
     ```go
     psql.Insert(
       im.OnConflict("did").DoUpdate(
         im.SetCol("dname").To(im.Excluded("dname")),
       ),
     )     
     // Renders: EXCLUDED.dname
     ```
     
     ```go
     im.OnConflict("id").DoUpdate(
         im.SetExcluded("title"),
         im.Where(im.Excluded("id").EQ(psql.Arg(1))),
     ),
     // Renders: ON CONFLICT (id) DO UPDATE SET title = EXCLUDED.title WHERE EXCLUDED.id = $1
     ```

2. **Refactoring**:
   - Updated PostgreSQL and SQLite helpers to use `Excluded` and `expr.Join` with expr.NoSep.


stephenafamo:
I'm a bit conflicted about `expr.Glue`

While I see the utility, the SQL produced without it is still technically correct, so I wonder if it is worth increasing the API surface of Bob.

atzedus:
Honestly, I also had doubts about whether this separate method (expr.Glue) is even needed. Currently it's only necessary in this one place and can generally be replaced by simply calling expr.Join with the appropriate parameters (e.g., with an empty separator).
Since you also have doubts, I think you're right — this is indeed premature API expansion. I agree it's better to remove it for now and keep the API minimal. We can always revisit this if a broader use case emerges later.

PR #689, opened 2026-05-18
Add multi-source FROM/USING support for PostgreSQL UPDATE/DELETE

atzedus:
This PR adds support for multiple `from_item` sources in PostgreSQL `UPDATE ... FROM` and `DELETE ... USING` via new append-style modifiers, without breaking existing API.

#### Features

- New: `um.FromAppend(...)` and `dm.UsingAppend(...)` allow adding multiple sources to the `FROM`/`USING` clause.
- Existing `um.From(...)` and `dm.Using(...)` keep their override (last call wins) behavior.
- Mixed usage is supported: calling `From/Using` after append resets the list; calling append after `From/Using` extends it.

#### Examples

```go
// Multiple FROM sources in UPDATE
psql.Update(
    um.Table("t"),
    um.SetCol("x").ToArg(1),
    um.FromAppend("a").As("a1"),
    um.FromAppend("b"),
)
// SQL: UPDATE t SET x = $1 FROM a AS a1, b

// Mixed: override after append
psql.Update(
    um.Table("t"),
    um.FromAppend("a"),
    um.From("b"),
)
// SQL: UPDATE t FROM b

// Multiple USING sources in DELETE
psql.Delete(
    dm.From("t"),
    dm.UsingAppend("a"),
    dm.UsingAppend("b").Only(),
)
// SQL: DELETE FROM t USING a, ONLY b
```

All existing usages of `From`/`Using` remain fully compatible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added PostgreSQL UPDATE/DELETE `um.WhereCurrentOf(cursor)` modifier to render `WHERE CURRENT OF <cursor>`. (thanks @atzedus)
 - Added `expr.NoSep` sentinel to join expressions without any separator. The existing `expr.Join` default (space) is unchanged. (thanks @atzedus)
 - Added `im.Excluded(column)` helper (PostgreSQL and SQLite) for `ON CONFLICT DO UPDATE` assignments — renders as `EXCLUDED."col"` with no extra space, reused internally by `im.SetExcluded(...)` and available for direct use in `im.SetCol(...).To(...)`. (thanks @atzedus)
+- Added PostgreSQL `psql.TableFunctions(funcs)` and `sm.FromFunction` / `um.FromFunction` / `dm.UsingFunction` helpers that return `bob.Expression` for table-function `from_item` sources (`ROWS FROM (...)` when multiple functions are given). (thanks @atzedus)
+- Added PostgreSQL `um.From(table, joins...)` and `dm.Using(table, joins...)` variadic join arguments so each appended `from_item` can include inline `INNER JOIN`, `LEFT JOIN`, `CROSS JOIN`, etc. (`JoinChain` builders). (thanks @atzedus)
+- Added `bobgen-psql` support for multiple `UPDATE ... FROM` / `DELETE ... USING` sources: sql-to-code emits `AppendTableRef(...)` per parsed `from_item` and resolves column sources from every item. (thanks @atzedus)
 
 ### Changed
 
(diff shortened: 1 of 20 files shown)


stephenafamo:
I think I would rather change `From` to take multiple items and append by default than add a separate `FromAppend`

atzedus:
Yes, I've also considered this approach, but in that case, we would need to do something like:

```go
um.From(
  um.FromTable("test1").As("t1"),
  um.FromTable("test2").As("t2"),
)
```

To preserve the ability to use `FromChain`. The `um.From()` method itself would need to be changed to not return `FromChain`.

In any case, these are breaking changes.

The situation requiring multiple tables is relatively rare, and in most cases, the single `um.From()` variant will be more commonly needed.

However, if we change the signature as I proposed above, it would look somewhat inconvenient from an API perspective - something like:

```go
psql.Update(
   ...
   um.From(um.FromTable("test1").As("t1"))
   ...
)
```

Regarding this functionality, I've been thinking about a few alternative approaches we could consider:

1. **Change the default behavior of `From`/`Using`**: Instead of adding `FromAppend`/`UsingAppend`, we could modify `From`/`Using` so that each subsequent call appends a new table rather than replacing the previous one. This wouldn't increase the API surface, but it would be a significant breaking change for users relying on the current "last call wins" behavior.

2. **Global package configuration**: Introduce a global setting for the `psql` package (e.g., via a package-level variable) that toggles this behavior, defaulting to the current non-breaking semantics. This avoids breaking changes but adds hidden state, which might reduce clarity and testability.

3. **Add a second parameter to `From`/`Using`**: For example, `From(table, mode)` where `mode` is a constant like `FromSet` or `FromAppend`. This keeps the method count low but still constitutes a breaking change due to the altered signature.

4. **Make the mode parameter optional via `OptionFunc`**: While flexible, this feels overly complex and less transparent for the API consumer.

5. **Add an `Append()` method to `FromChain`/`UsingChain`**: This would allow chaining append operations, but again, the API intent becomes less obvious compared to explicit method names.

After working through this PR and weighing these options, I've come to the conclusion that the current approach - introducing explicit `FromAppend`/`UsingAppend` methods - is the most transparent, predictable, and non-breaking solution from an API design perspective.

That said, I'm open to other ideas if you has suggestions on how else this could be implemented, or if we're comfortable with the proposed approach as-is.

stephenafamo:
Personally, I prefer option 1

In my opinon, Bob should try to behave how the SQL will behave. If the query takes multiple FROM, then using multiple FROM query mods should add multiple FROM entries.

Yes it is a big breaking change, but as you've said, it is rare. And even today, I would hope no one is explicitly depending on the "last call wins" attribute.

atzedus:
I like this option the most as well, if not for the breaking changes — but if we decide to go this route, I think we should indeed proceed in this direction.

However, after reviewing the current changes once more, I noticed that `FromAppend`/`UsingAppend` only add the table itself as a "base" source. Yet, according to the SQL specification, each `from_item` can contain additional expressions like `LEFT JOIN`, `ON` conditions, etc. In the current implementation, modifiers like `LeftJoin`, `InnerJoin`, and others are attached to the single table specified in `From` as independent mods.

To properly support multiple `FROM`/`USING` sources, a deeper rework would be required: each call to `From`/`Using` should also independently allow attaching its own set of `JOIN` clauses. This significantly increases the scope of changes.

Given all of this, I think we can close the current PR for now. I'll try to revisit this problem later with a more comprehensive approach.

atzedus:
On second thought, I'm leaning toward a slightly different API concept that could address both the multi-source and per-source joins concerns:

```go
func From(table any, mods ... dialect.JoinChain[*dialect.UpdateQuery])
func Using(table any, mods ... dialect.JoinChain[*dialect.DeleteQuery])
```

The idea is to encapsulate the table + its associated joins/conditions into a composable `JoinChain` modifier. This way, each `From`/`Using` call could accept its own self-contained chain of joins, making the API both flexible and semantically clear:

```go
// Example (conceptual):
psql.Update(
  um.Table("example"),
  um.From("orders",
      um.InnerJoin("users").On("orders.user_id = users.id"),
      um.LeftJoin("payments").On("orders.id = payments.order_id"),
  um.From("warehouses",
      um.InnerJoin("inventory").On("w.id = inventory.warehouse_id"),
  ).As("w"),
)
```

I think it's worth moving in this direction, ideally together with changing the default behavior of `From`/`Using` to support multiple sources natively. This would give us a clean, SQL-aligned API without requiring separate `*Append` variants.

What do you think?

stephenafamo:
> I think it's worth moving in this direction, ideally together with changing the default behavior of `From`/`Using` to support multiple sources natively. This would give us a clean, SQL-aligned API without requiring separate `*Append` variants.
> 
> What do you think?

I love it 😁

PR #691, opened 2026-05-21
feat(dialect): add TableName, Schema and ColumnsExpr to View types

atzedus:
## Summary

- Add `TableName()` to `mysql.View`, `psql.View`, and `sqlite.View` so views expose the underlying name the same way tables do in generated models and query helpers.
- Add `Schema()` to `psql.View`, and `sqlite.View` so views expose the underlying name the same way tables do in generated models and query helpers.
- Add `ColumnsExpr()` to `psql.View` for parity with MySQL and SQLite views (those dialects already had it).

## Motivation

`View` types were missing a few accessors that make working with views consistent across dialects and with other ORM helpers. `TableName()` makes the view name available without going through `Name()` / quoting, and `ColumnsExpr()` on PostgreSQL closes a small API gap compared to the other dialects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added `TableName()` method to dialect `View` types (MySQL, PostgreSQL, SQLite). (thanks @atzedus)
+- Added `ColumnsExpr()` method on `psql.View` for parity with MySQL and SQLite views. (thanks @atzedus)
+
 ## [v0.44.0] - 2026-05-21
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `Schema()` method to dialect `View` types (PostgreSQL, SQLite). (thanks @atzedus)
 - Added `TableName()` method to dialect `View` types (MySQL, PostgreSQL, SQLite). (thanks @atzedus)
 - Added `ColumnsExpr()` method on `psql.View` for parity with MySQL and SQLite views. (thanks @atzedus)
 
--- a/dialect/mysql/view.go
+++ b/dialect/mysql/view.go
@@ -59,7 +59,12 @@ func (v *View[T, Tslice, C]) Alias() string {
 	return v.alias
 }
 
-// Returns a column list
+// TableName returns the table name
+func (v *View[T, Tslice, C]) TableName() string {
+	return v.name
+}
+
+// ColumnsExpr returns a column list
 func (v *View[T, Tslice, C]) ColumnsExpr() expr.ColumnsExpr {
 	// get the schema
 	return v.allCols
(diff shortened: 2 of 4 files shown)


atzedus:
Thank you so much for the latest release! I’ve already started using the new features, and amidst all the great improvements, I realized one small detail was missed. This PR adds that final piece to complete the recent updates.

I would greatly appreciate it if you could cut a patch release (`v0.44.1`) after merging this PR to include these final changes.

Thanks again for your continued work on this project!

@stephenafamo

atzedus:
While working with the new `View` API, I noticed that the method names `Name()` and `NameAs()` might be slightly ambiguous — they return a table expression (with optional schema), not just a simple name string. This could potentially cause some confusion for new users.

I've been thinking about introducing clearer alternatives:

```go
func (v *View[T, Tslice, C]) TableExpr() Expression {
	// schema is not empty, never override
	if v.schema != "" {
		return Quote(v.schema, v.name)
	}
	return Expression{}.New(orm.SchemaTable(v.name))
}

func (v *View[T, Tslice, C]) TableExprAs() bob.Expression {
	return v.TableExpr().As(v.alias)
}
```

And then mark the current `Name()` / `NameAs()` methods as `// deprecated` (keeping them for backward compatibility, of course).

That said, I'm not entirely sure this change is necessary — it's more of a "nice-to-have" for API clarity. If you think this is worth doing, I'd be happy to open a small separate PR with the new methods + deprecation notices. If not, no worries at all! Just wanted to float the idea. 😊

stephenafamo:
> it's more of a "nice-to-have" for API clarity.

Yes. Naming is hard...  
Because with your change, then I worry that "Table" is technically not correct for views. I also wonder if perhaps `Expr/ExprAs` or `NameExpr/NameAsExpr`

Then again, considering that it returns an expression, perhaps the change is not needed since it cannot be accidentally used as a raw string.

atzedus:
> Yes. Naming is hard... Because with your change, then I worry that "Table" is technically not correct for views. I also wonder if perhaps `Expr/ExprAs` or `NameExpr/NameAsExpr`

When choosing `TableName()`, I was primarily guided by consistency with existing constructor signatures in the codebase, such as:

```go
func NewView[T any, C bob.Expression](tableName string, columns C)
func NewTablex[T any, Tslice ~[]T, Tset setter[T], C bob.Expression](tableName string, columns C, uniques ...[]string)
```
Also, `Name()` was already occupied by the existing method that returns an expression, so I wanted to avoid overloading it with a different return type.

From the perspective of generated code, these methods will be used as `models.Users.TableName()` / `models.Users.Schema()`, which felt intuitive and aligned with how tables expose their metadata.

That said, I completely agree that the *ideal* API might look more like for view:

```go
Name() string           // raw table/view name
NameExpr() Expression   // quoted expression, with optional schema
NameAsExpr() Expression // same, but with alias applied
```

This would make the distinction between raw identifiers and SQL expressions more explicit. However, introducing this now would be a breaking change for anyone already using `Name()`/`NameAs()`, and given that the return type is already an `Expression` (so it can't be accidentally used as a raw string), I'm inclined to agree that the current trade-off is acceptable for now.

stephenafamo:
> That said, I completely agree that the _ideal_ API might look more like for view:
> 
> ```go
> Name() string           // raw table/view name
> NameExpr() Expression   // quoted expression, with optional schema
> NameAsExpr() Expression // same, but with alias applied
> ```
> 
> This would make the distinction between raw identifiers and SQL expressions more explicit.

I agree and I think it makes sense to make this breaking change now.

Yes it is never fun to introduce breaking changes, but:
1. This is still pre 1.0
1. Better to do it now than later or just stick with a poor name
1. This is a search/replace upgrade so would take a few minutes at most

PR #692, opened 2026-05-22
feat(dialect): add NameExpr/NameAsExpr and fix schema column qualifiers

atzedus:
This PR completes the View/Table naming API started in #691 and fixes a codegen bug where generated column prefixes could disagree with the table alias used in SQL.

- **BREAKING:** `View` / `Table` `Name()` now returns the bare table or view name as a `string`. SQL builders should use `NameExpr()` (formerly `Name()`) and `NameAsExpr()` (formerly `NameAs()`).
- **Added:** `Schema()` on PostgreSQL and SQLite `View` types (MySQL is unchanged — no schema/database qualification on models).
- **Fixed:** Introduced `tableColumnAlias` in codegen so `build*Columns`, `pkEQ`, and `pkIN` use the same qualifier as `View.Alias()` when `schema` is set at construction (`schema.table` instead of a bare `table.Key` such as `users`).
- Updated codegen templates, `orm.Preload`, factory test templates, CHANGELOG, and view/table docs.

## Codegen fix: column qualifiers vs table alias

The bug is not about renaming columns or tables — it is about **which prefix** codegen attaches to the same `users.id` when the model is built with a schema.

Assume the same table everywhere:

| Field | Value |
|-------|--------|
| Table | `users` |
| Column | `id` |
| Generated model | `models.Users` |

Only the **schema** at construction differs.

### Case A — `shared_schema` (schema empty at codegen)

Codegen emits:

```go
var Users = psql.NewTablex(..., "", "users", buildUsersColumns("users"), ...)
```

Runtime:

| | Value |
|--|--------|
| `Users.Schema()` | `""` |
| `Users.Name()` | `"users"` |
| `Users.Alias()` | `"users"` |
| `Users.Columns.ID` | `"users"."id"` |
| `Users.NameAsExpr()` | `"users" AS "users"` |

Manual query:

```go
psql.Select(
    sm.From(models.Users.NameAsExpr()),
    sm.Where(models.Users.Columns.ID.EQ(1)),
)
```

```sql
SELECT "users"."id"
FROM "users" AS "users"
WHERE ("users"."id" = $1)
```

Prefix `users` is consistent everywhere. **No bug** — `table.Key` (`"users"`) matches `Alias()`.

---

### Case B — non-shared schema (schema `billing` at codegen)

Codegen emits:

```go
var Users = psql.NewTablex(..., "billing", "users", buildUsersColumns(...), ...)
```

Runtime (dialect):

| | Value |
|--|--------|
| `Users.Schema()` | `"billing"` |
| `Users.Name()` | `"users"` |
| `Users.Alias()` | `"billing.users"` |
| `Users.NameAsExpr()` | `"billing"."users" AS "billing.users"` |
| `Users.Query()` columns (via `View.ColumnsExpr`) | `"billing.users"."id"` |

**Before this PR** — codegen passed `table.Key` (`"users"`) into `buildUsersColumns` and `pkEQ`:

| | Value |
|--|--------|
| `buildUsersColumns("users")` | `"users"."id"` |
| `pkEQ()` | `Quote("users", "id")` |

Manual query with the **same** `models.Users` and **same** `id`:

```go
psql.Select(
    sm.From(models.Users.NameAsExpr()),
    sm.Where(models.Users.Columns.ID.EQ(1)),
)
```

```sql
SELECT "users"."id"
FROM "billing"."users" AS "billing.users"
WHERE ("users"."id" = $1)
```

`Users.Query()` alone looked fine (it uses `View.ColumnsExpr` with parent `"billing.users"`), but `Users.Columns.ID` and `pkEQ()` pointed at `"users"."id"` while the SQL alias was `"billing.users"`.

**After this PR** — `tableColumnAlias("billing", "users", "users")` → `"billing.users"`:

```go
var Users = psql.NewTablex(..., "billing", "users",
    buildUsersColumns("billing.users"), ...)
```

| | Value |
|--|--------|
| `Users.Columns.ID` | `"billing.users"."id"` |
| `pkEQ()` | `Quote("billing.users", "id")` |

Same manual query:

```sql
SELECT "billing.users"."id"
FROM "billing"."users" AS "billing.users"
WHERE ("billing.users"."id" = $1)
```

Same table `users`, same column `id` — only the schema argument changes the prefix, and codegen now matches `View.Alias()`.

### Migration

| Before (v0.44.0) | After |
|------------------|--------|
| `Users.Name()` in `sm.From`, joins, etc. | `Users.NameExpr()` |
| `Users.NameAs()` | `Users.NameAsExpr()` |
| Need qualified name as `Expression` | `NameExpr()` / `NameAsExpr()` |
| Need unqualified name string | `Users.Name()` |
| Schema on psql/sqlite | `Users.Schema()` |

Regenerate models after upgrading bobgen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `Schema()` method to dialect `View` types (PostgreSQL, SQLite). (thanks @atzedus)
-- Added `TableName()` method to dialect `View` types (MySQL, PostgreSQL, SQLite). (thanks @atzedus)
 - Added `ColumnsExpr()` method on `psql.View` for parity with MySQL and SQLite views. (thanks @atzedus)
+
+### Changed
+
+- **BREAKING:** `View` / `Table` `Name()` now returns the bare table (or view) name as `string`. In v0.44.0, `Name()` returned a dialect `Expression` (qualified table reference for SQL). That role moved to new methods:
+  - `NameExpr()` — what `Name()` did in v0.44.0 (on PostgreSQL and SQLite, respects `UseSchema` when schema was generated empty)
+  - `NameAsExpr()` — what `NameAs()` did in v0.44.0 (name plus table alias)
+  Update manual call sites, codegen templates, and generated models: `Users.Name()` in SQL builders → `Users.NameExpr()`, `Users.NameAs()` → `Users.NameAsExpr()`; use `Users.Name()` when you need the unqualified name string. On PostgreSQL and SQLite, `Schema()` still returns the schema when set.
+- Codegen templates and `orm.Preload` now use `NameExpr()` / `NameAsExpr()` instead of `Name()` / `NameAs()`. (thanks @atzedus)
+
+### Fixed
+
+- Fixed generated models using the wrong column qualifier when a table was constructed with a non-empty schema. Codegen previously passed `table.Key` into `build*Columns` and `pkEQ` / `pkIN`, but dialect `View` / `Table` types use `schema.table` as `Alias()` whenever `schema` is set at construction. That mismatch produced SQL such as `` `users`.`id` `` in `WHERE` while `FROM` used `` `myapp`.`users` AS `myapp.users` `` (columns and primary-key expressions did not match the table alias from `NameAsExpr()`). Templates now use a `tableColumnAlias` helper: `schema.name` when `schema` is set, otherwise `table.Key`. (thanks @atzedus)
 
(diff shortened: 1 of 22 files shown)


stephenafamo:
This looks good.

I wonder if there is a way to prevent compilation if the models are generated with a version of Bob that is different from the runtime.

atzedus:
Good question. I think, that exact semver equality at compile time isn’t really doable.

What we can do is a codegen API version: like this does protobuf - https://pkg.go.dev/google.golang.org/protobuf/runtime/protoimpl#EnforceVersion.

This PR already forces a compile break for the Name() → NameExpr() API change.

PR #696, opened 2026-05-26
fix(psql): CRITICAL. Merge existing joins in SelectQuery.AppendTableRef

atzedus:
Fixes PostgreSQL `SELECT` queries where join mods run before `sm.From`, or where a later `sm.From` should keep joins already attached to the embedded `FROM` `from_item`.

This fix affects both standard sm helpers like sm.InnerJoin and sm.LeftJoin, as well as codegen-generated ones (e.g., Preloads).

`SelectQuery.AppendTableRef` (used by `sm.From` / `FromChain`) used to assign `s.TableRef = ref` wholesale, so any joins added earlier via `sm.CrossJoin`, `sm.InnerJoin`, etc. were dropped.

## What broke

After `sm.From` was wired through `AppendTableRef` (replacing the old `SetTable` / `SetTableAlias` path), the primary `FROM` table ref was always replaced as a whole. Join mods only append to `TableRef.Joins`; they do not set `Expression`. If `sm.From` ran after them, those joins were lost.

### Join before `From` (mod order)

```go
psql.Select(
    sm.Columns("id"),
    sm.CrossJoin("orders"),
    sm.From("users"),
)
```

**Expected:** `SELECT id FROM users CROSS JOIN orders`  
**Got:** `SELECT id FROM users`

### Join after first `From`, then second `From`

```go
psql.Select(
    sm.Columns("id"),
    sm.From("users"),
    sm.CrossJoin("events"),
    sm.From("orders"),
)
```

**Expected:** `SELECT id FROM orders CROSS JOIN events`  
**Got:** `SELECT id FROM orders`

The usual order (`sm.From` then `sm.CrossJoin`) still worked; only reordering or a second `From` exposed the bug.

## Fix

`AppendTableRef` now merges existing `Joins` into the incoming ref before assignment:

```go
if len(s.TableRef.Joins) > 0 {
    ref.Joins = append(s.TableRef.Joins, ref.Joins...)
}
s.TableRef = ref
```

Primary table fields (`Expression`, alias, `ONLY`, `TABLESAMPLE`, etc.) still come from the latest `sm.From`; only `Joins` are preserved/merged.

New cases for unit-test: `from multiple tables cross join unordered calls`, `from keeps existing joins`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed PostgreSQL `sm.From` clearing joins already on the `FROM` from_item (e.g. `sm.CrossJoin` before `sm.From`) because `SelectQuery.AppendTableRef` replaced the entire embedded `FROM` table ref. Existing joins are merged into the new primary table ref. (thanks @atzedus)
 - Fixed `FOR UPDATE` / `FOR SHARE` `OF` table lists: pass `psql.Quote(...)` / `mysql.Quote(...)` so names with spaces or qualification render correctly ([#693](https://github.com/stephenafamo/bob/issues/693)). (thanks @atzedus)
 - Fixed PostgreSQL `MERGE` `INSERT` column lists (`mm.Columns`): names are written unquoted (not via `WriteQuoted` / `expr.Quote`), so PostgreSQL still folds unquoted identifiers to lowercase and `mm.Columns("Name")` matches a column created as `name`. Quoted identifiers are case-sensitive in PostgreSQL; auto-quoting broke that. Use `psql.Quote(...)` in value expressions when you need a delimited identifier. (thanks @atzedus)
 - Fixed `um.SetCols` / `im.SetCols` / `mm.SetCols` tuple-assignment column lists the same way (unquoted `[]string` names on the left-hand side of `(cols...) = ...`). (thanks @atzedus)
--- a/dialect/psql/dialect/select.go
+++ b/dialect/psql/dialect/select.go
@@ -36,7 +36,11 @@ type SelectQuery struct {
 	CombinedOffset clause.Offset
 }
 
+// AppendTableRef sets the primary FROM from_item on the query.
 func (s *SelectQuery) AppendTableRef(ref clause.TableRef) {
+	if len(s.TableRef.Joins) > 0 {
+		ref.Joins = append(s.TableRef.Joins, ref.Joins...)
+	}
 	s.TableRef = ref
 }
 
--- a/dialect/psql/select_test.go
+++ b/dialect/psql/select_test.go
@@ -39,6 +39,16 @@ func TestSelect(t *testing.T) {
 				sm.From("orders"),
 			),
 		},
+		"from keeps existing joins": {
+			Doc:         "A later From replaces the primary table but keeps joins already on the from_item",
+			ExpectedSQL: "SELECT id FROM orders CROSS JOIN events",
+			Query: psql.Select(
+				sm.Columns("id"),
+				sm.From("users"),
+				sm.CrossJoin("events"),
+				sm.From("orders"),
+			),
+		},
 		"from multiple tables cross join": {
 			Doc:         "Multiple tables in FROM via CROSS JOIN on the primary from_item",
 			ExpectedSQL: "SELECT id FROM users CROSS JOIN orders",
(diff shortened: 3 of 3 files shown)


atzedus:
@stephenafamo Would it be possible to publish a patch release v0.44.1 once this is merged? This fixes a pretty critical bug that could silently break queries for many users. It's especially likely to surface when using Preload (which was my case), since in hand-written queries the pattern of calling From after InnerJoin/LeftJoin is probably rare.

atzedus:
While this fix makes perfect sense given the current behavior and the `Preload` use case, it brings up an interesting API design discussion.

**1. The "Explicit Replacement" perspective**
Consider a manual usage pattern like this:
```go
psql.Select(
    sm.Columns("id"),
    sm.From("users"),
    sm.CrossJoin("events"),
    sm.From("orders"),
)
```
What is the user's actual intent here? Do they want to replace just the previously defined `FROM` table, or do they want to replace the *entire* `FROM` clause, discarding all previously added joins? 
From this angle, a more explicit API design would be similar to what is currently done for `UPDATE` queries: `um.From(table any, joins ...dialect.JoinChain[*dialect.UpdateQuery])`. That signature makes it completely obvious that the entire `FROM` block and its associated joins are being replaced wholesale.

**2. The Codegen / Preload perspective**
On the flip side, look at how it works with code generation (e.g., Preloads):
```go
psql.Select(
    sm.Column(....),
    models.Preload.OrderItems,
    sm.From(models.Orders.Name()),
)
```
Here, the current API feels much more intuitive. The developer doesn't need to know that `Preload` injects a `LEFT JOIN` under the hood. They just state their intent: "preload `order_items` for this table", and they shouldn't have to care about the internal SQL mechanics.

**3. Composability and `sql-to-code`**
Furthermore, having independent mods for `From`, `InnerJoin`, `LeftJoin`, etc., is highly convenient. It allows for the "fine-grained" composition of base query mods, which is especially powerful for `sql-to-code` workflows and building reusable query fragments.

**4. Is it actually a bug?**
Philosophically, one could argue this isn't even a bug if `sm.From` is strictly meant to overwrite. But practically, given features like `Preload` and the library's historical behavior, it definitely behaves like a critical bug in the wild.

**5. The Main Question: Should we change the API design?**
This leads to the main question: should we eventually update `sm.From` to match the signature of `um.From`, i.e., `From(table any, joins ...dialect.JoinChain[*dialect.SelectQuery])`?

*   **Pros:** It adds clarity and explicit control. If joins are explicitly passed to `sm.From`, it overwrites any previously added joins. If no joins are passed, it leaves the existing joins intact.
*   **Cons:** It might confuse users. If `sm.From` accepts joins directly, it raises the question: why do we still need separate `sm.InnerJoin` and `sm.LeftJoin` mods?

Would love to hear your thoughts on this, @stephenafamo. Do you think the current fix is the best path forward, or should we consider an API change in a future major/minor release?

stephenafamo:
Considering that joins are inherently part of the `from_item`, I think **OPTION 5** is the most intuitive fix.

I don't love the breaking change, but I feel like it is better long term since it avoids any surprises.

stephenafamo:
> @stephenafamo Would it be possible to publish a patch release v0.44.1 once this is merged?

Sure

atzedus:
> Considering that joins are inherently part of the `from_item`, I think **OPTION 5** is the most intuitive fix.
> 
> I don't love the breaking change, but I feel like it is better long term since it avoids any surprises.

I generally like this approach and think it's a good direction to move in. I'll analyze this further and review the relevant use cases.
Preliminarily, I think we should consider removing InnerJoin and similar mods as separate modifiers. This would definitely affect the Preload mechanism at minimum. I'll follow up on this in a separate issue later.
But for now, this is a bug that I'd really like to get fixed as soon as possible :)

stephenafamo:
> Preliminarily, I think we should consider removing InnerJoin and similar mods as separate modifiers.

Yes I agree.

> This would definitely affect the Preload mechanism at minimum. 

Yes it will. Not sure how it will have to look.

Preload does 3 things:

1. Add some columns to the select list
2. Attach one or more joins
3. Add a loader to extract the final result

We can't then move preloading to simply a "JoinMod" since that only covers 2. Perhaps adding the requisite "joins" as part of `bob.Load` would work 🤔

atzedus:
Thanks, I'll take a look at this in a future PR.

stephenafamo:
Is it okay to close this?

atzedus:
> Is it okay to close this?

Could you please reconsider merging this PR? This fixes a critical regression introduced in v0.44.0 that breaks query generation when using `Preload`, and it's important to get it into a patch release (v0.44.1) to prevent silent failures for users.

I fully agree that the broader API discussion around `sm.From` and join handling (option 5) is valuable — but I believe that should be handled in a separate, follow-up PR. There's no need to block this urgent bugfix on a larger redesign; we can iterate on the API in a future minor/major release.

Ideally, this could be merged together with #697 before cutting v0.44.1, since both address issues from 0.44.0. But if #697 needs more time, I'm fine with releasing this fix first.

Issue #695, opened 2026-05-25
Inconsistent auto-quoting of column identifiers (string params)


atzedus:
### Problem

In PostgreSQL, unquoted identifiers fold to lowercase; quoted identifiers are case-sensitive. Several helpers take column/alias names as plain `string` / `...string` and **always** render them via `WriteQuoted` or `expr.Quote`. Passing `"Name"` then produces `"Name"` in SQL, which may not match a column stored as `name`.

The same kind of identifier is already rendered **without** auto-quotes in other helpers (e.g. `SetCols`, `OnConflict("id")`, MERGE `Columns` after recent fixes). That makes the API inconsistent.

**Example (aliases + references):**

```go
sm.From("accounts").As("TaBLe1")
// → FROM accounts AS "TaBLe1"

psql.Quote("table1", "id")  // → "table1"."id"
// does not refer to alias "TaBLe1"; wrong identifier if you expected folding to table1
```

Use the same spelling everywhere (`Quote("TaBLe1", "id")`) or stop auto-quoting `As("TaBLe1")` so the alias folds to `table1`.

### Proposed direction

Render plain `string` identifiers **unquoted** in the APIs below. Use `psql.Quote` / `mysql.Quote` / `sqlite.Quote` when quoting is required. **Breaking:** call sites that relied on implicit quoting must pass `Quote(...)` explicitly.

### APIs that auto-quote today (should be aligned)

| API | Dialects | Mechanism |
|-----|----------|-------------|
| `SetCol(...).To` / `.ToArg` | psql `um`/`im`/`mm`, sqlite `um`/`im`, mysql `um` | `mods.Set` → `expr.Quote` |
| `SetExcluded` / `Excluded` | psql, sqlite `im` | `expr.Quote` on SET column and `EXCLUDED.col` |
| `Into` / `IntoAs` (column list) | psql, sqlite, mysql `im` | `clause.TableRef.Columns` → `WriteQuoted` |
| `From` / `Join` / `CrossJoin` `.As(alias)` or `.As(alias, columns...)` | psql, mysql, sqlite | `TableRef` alias / columns → `WriteQuoted` |
| `Join.Using` / `UsingAs` | psql, mysql, sqlite | `expr.Quote` / `WriteQuoted` |
| `TableAs` / `FromAs` / `IntoAs` (alias) | psql | `TableRef.Alias` → `WriteQuoted` |
[...]

PR #697, opened 2026-05-26
fix: quote SQL identifiers across dialects

atzedus:
As discussed in #695.

## Summary

This PR improves identifier handling and `SET` assignment APIs across dialects.

## What changed

- Fixed identifier quoting in shared clause builders by using internal `QuoteIdentifiers` for:
  - CTE names/columns
  - window names
  - `ON CONFLICT ON CONSTRAINT`
  - `WHERE CURRENT OF`
  - partition/index-hint related identifier lists
- Updated `mods.Set` internals (`Col any`) so assignment LHS is rendered via `bob.Express`.
- Kept `SetCol` / `UpdateCol` APIs as `string`-based for simple single-column names, with automatic quoting via `expr.Quote`.
- Added/standardized `SetExpr(...)` for qualified or arbitrary LHS expressions in `SET` / `DO UPDATE SET`.
- Clarified that `Set(...)` should be used for grouped/qualified assignment expressions.
- Changed function rendering so function names go through `F(name any)` + `bob.Express` (instead of raw string-only names).

## SetExpr details

- `SetCol("col")` / `UpdateCol("col")`: for simple single identifiers (auto-quoted).
- `SetExpr(expr)`: for qualified or custom LHS expressions, e.g. `dialect.Quote("t", "col")`.
- Both forms can be used directly (`SetCol(...).To(...)`, `SetExpr(...).To(...)`) and also passed into grouped `Set(...)` mods when composing multiple assignments.

## Validation and docs

- Updated tests for MySQL, PostgreSQL, and SQLite to cover new quoting and `SetExpr` behavior.
- Updated quoting docs (`website/docs/query-builder/quotes.md`) and MySQL update examples.


atzedus:
@stephenafamo `go test ./...` success locally.

stephenafamo:
Yeah, the test passes here too. Fails when trying to upload the coverage to coveralls

atzedus:
@stephenafamo PR ready for review

atzedus:
While working on this PR I noticed that dialect unit tests rely on a `formatter` passed to `testutils.RunTests`. As I understand it, this was originally added to (a) verify that generated SQL is syntactically valid and (b) make assertions less sensitive to whitespace/formatting differences.

However, I think it also hides some real issues:

- The fallback `testutils.Clean()` strips all identifier quotes, collapses nested parentheses (via `( (` → `( `), and removes parentheses around simple `ON`/`WHERE`/`AND` clauses.
- The dialect-specific formatters (pg_query parse/deparse, ANTLR `GetText()`) further normalize the SQL before comparison.

One example from this PR: extra parentheses around subqueries were not obvious until I stopped relying on normalization and looked at the raw generated SQL. I suspect there may be other cases (extra/missing parens, quoting differences) that the formatter currently masks.

If you're okay with it, I'd like to follow up after this PR merges with a separate PR that removes `formatter` from dialect unit tests and compares generated SQL more literally (likely keeping only minimal normalization such as whitespace, if needed). Tests may become a bit more verbose, and we'd lose the implicit "generated SQL must parse" check unless we add that explicitly where it still adds value — but I think stricter assertions would help catch regressions in SQL generation earlier.

Does that approach sound reasonable to you, or are there cases where you still consider the formatter important?

@stephenafamo

stephenafamo:
The purpose of the formatter is not to fail semantically equivalent queries. For me the "gold standard" is pg query's parse/deparse 

Considering that the tests are used to build the examples, this is especially useful.

For parentheses, this makes sense since I don't think a query should fail an "equivalence" check for adding an additional set of parentheses.

Now if this can be tweaked to not reject equivalent code but perhaps be stricter.. I'm all for it.

On the other hand, perhaps a different set of focused tests can be used to check for extra parentheses.

atzedus:
> For parentheses, this makes sense since I don't think a query should fail an "equivalence" check for adding an additional set of parentheses.

Thanks, that makes sense now. It's a reasonable approach - I'll think about how to preserve the current implementation while also adding checks for parentheses, quotes, and similar formatting details, without overcomplicating the tests.

PR #699, opened 2026-05-28
fix: iterate over all schema constraint errors not just the first

### LINKED ISSUE #698: DB Has multiple constraints, only 1 constraint error is generated
My Schema name is the same as my database name, causing bob to not properly generate the schema prefix. To work around this, I have  `shared_schema: "__none__"` defined in my `bobgen.yaml` to properly generate the prefix 

This however, causes an issue in this file: https://github.com/stephenafamo/bob/blob/main/gen/bobgen-psql/driver/keys.go line 101.
```
case "u":
	ret.Uniques[key] = append(ret.Uniques[c.Table], drivers.Constraint[any]{
	Name: c.Name,
	Columns: c.Columns,
	Comment: c.Comment.String,
})
```

The issue is in my case, `ret.Uniques[c.Table]` is nil (since it will be prefix.tableName, not tableName). Therefore, if there are multiple constraints only the last constraint is left remaining in the slice. E.g.

`slice = append (nil, constraint1)
`
`slice = append(nil, constraint2)`

This leave `slice == [constraint2]` instead of `slice == [constraint1, constraint2]`

And in my generated code:

	ErrUniqueConstraintKey: &UniqueConstraintError{
		schema:  "schemaName",
		table:   "myTable",
		columns: []string{"affected_column"},
		s:       "constraint2",
	}

This is likely a one-line fix:

```
ret.Uniques[key] = append(ret.Uniques[key], drivers.Constraint[any]{ ... })
```

MD-Mushfiqur123:
Fixes #698

In `gen/bobgen-psql/driver/keys.go`, the unique constraints were being appended using `c.Table` instead of `key` as the lookup key. When a table belongs to a non-shared schema, `key` includes the schema prefix while `c.Table` does not. This caused each subsequent unique constraint to append to a nil slice (since `ret.Uniques[c.Table]` is empty) instead of the existing slice at `ret.Uniques[key]`, effectively overwriting the previous constraint.

Only the last unique constraint was retained, meaning constraint errors would only ever detect at most one constraint violation.

### DIFF
--- a/gen/bobgen-psql/driver/keys.go
+++ b/gen/bobgen-psql/driver/keys.go
@@ -99,7 +99,7 @@ func (d *driver) Constraints(ctx context.Context, _ drivers.ColumnFilter) (drive
 			}
 
 		case "u":
-			ret.Uniques[key] = append(ret.Uniques[c.Table], drivers.Constraint[any]{
+			ret.Uniques[key] = append(ret.Uniques[key], drivers.Constraint[any]{
 				Name:    c.Name,
 				Columns: c.Columns,
 				Comment: c.Comment.String,


eaglehuntt:
@MD-Mushfiqur123 Hey, just following up on this since Stephen mentioned adding attribution for the original report in the changelog. No worries either way, but I wanted to check before the release goes out. Thanks again for the quick fix.

stephenafamo:
Also attributed to you @eaglehuntt in the CHANGELOG and release notes

PR #701, opened 2026-05-29
Render EQ in SET without comparison parens; omit redundant AS in NameAsExpr

atzedus:
`EQ` in SET clauses no longer gets comparison parentheses, and `NameAsExpr()` still drops a redundant `AS` when the alias matches the table name.

---

`EQ` in `WHERE` / `ON` is wrapped in parentheses by `expr.X` — that stays as-is for filters. The same expression in `SET` produced awkward SQL:

```sql
-- psql.Quote("kind").EQ(psql.Arg("Drama")) in um.Set(...)
UPDATE films SET ("kind" = $1)   -- extra parentheses
```

SET rendering now runs assignments through `expr.PrepareSetAssignments`, which rewrites `EQ` equalities so they render as `col = val` without that wrapper. `WHERE` and `ON` are unchanged:

```go
um.Set(
    psql.Quote("kind").EQ(psql.Arg("Drama")),
    psql.Quote("employees", "dept_id").EQ(psql.Quote("accounts", "dept_id")),
)
// UPDATE employees SET "kind" = $1, "employees"."dept_id" = "accounts"."dept_id"

um.Where(psql.Quote("kind").EQ(psql.Arg("Drama")))
// WHERE (kind = $2)
```

For a single column, `um.SetCol(...).To(...)` / `um.SetExpr(...).To(...)` remain the usual choice. `EQ` in `um.Set` / `im.Set` / `mm.Set` is useful when several assignments share one mod with an arbitrary LHS via `Quote`.

Also covered: `ON CONFLICT DO UPDATE SET`, PostgreSQL `MERGE ... UPDATE SET`, and MySQL `ON DUPLICATE KEY UPDATE`.

## `NameAsExpr()` and redundant alias

Previously `NameAsExpr()` always added `AS`, even when the alias matched the name:

```sql
FROM "users" AS "users"
```

Now `AS` is omitted when `alias == name` and no schema is set:

```sql
FROM "users"
```

If a schema was set at construction, the alias is still emitted (`"schema.table"`); that behavior is unchanged.

Affects `View` / `Table` on PostgreSQL, SQLite, and MySQL; tests and `website/docs/models/view.md` updated.


stephenafamo:
> Now `AS` is omitted when `alias == name` and no schema is set:

I like this. Thanks 👍🏾 



> Also added: `expr.Assign(left, right)` and `Chain.Assign` on all dialects.

Now this..... It feels too similar to `EQ` that I don't like it, feels like a potential source of confusion.  
Perhaps it would be better to modify `SET` where its args have a paren omitter or something. Not sure exactly how to implement this

atzedus:
> Now this..... It feels too similar to `EQ` that I don't like it, feels like a potential source of confusion.

I anticipated this outcome, which is exactly why this PR contains two separate commits — mostly to avoid merge conflicts in the changelog if we need to adjust one part independently :))

While implementing `Assign`, I actually debated with myself about whether it's truly necessary. Here are my arguments for why I believe it adds value:

1. **SQL semantics**: The `=` operator works in two distinct contexts — `Equal` in `WHERE` clauses and `Assign` in `SET` clauses. Having separate methods reflects this semantic difference.

2. **`EQ` can be misleading too**. Consider this example:

```go
um.Update(um.Set(
  psql.Quote("users", "name").EQ(psql.Arg("bob")), // looks like it should work, but it doesn't
))

// codegen
cols := dbmodels.Users.Columns
dbmodels.Users.Update(
  cols.Name.EQ(psql.Arg("bob")) // also doesn't work as intended
))
```

**What are the current alternatives for writing a correct query?**

```go
cols := dbmodels.Users.Columns
um.Update(um.Set(
  psql.Quote("users", "name").OP("=", psql.Arg("Bob")), // works, but doesn't look elegant
  expr.OP("=", psql.Quote("users", "name"), psql.Arg("Bob")), // not very readable
  um.SetExpr(psql.Quote("users", "name")).ToArg("bob"), // most acceptable, but verbose
))
```

**With the new `Assign`, the code becomes shorter and more readable:**

```go
cols := dbmodels.Users.Columns
um.Update(um.Set(
  psql.Quote("users", "name").Assign(psql.Arg("bob")),
  cols.Name.Assign(psql.Arg("bob")),
))
```

Ideally, `EQ` would intelligently omit the extra parentheses when used inside a `SET` clause, but I currently don't see a clean way to implement that context-aware behavior.

That said, if you still feel that `Assign` is unnecessary and we should hold off on it, just let me know — I'm happy to revert the commit with its implementation from this PR and keep only the `NameAsExpr` improvement.

stephenafamo:
> `Assign` is useful when several assignments are passed to one `um.Set` / `im.Set` / `mm.Set` with an arbitrary LHS via `Quote`.

Why not just pass multiple `SetCol` in place of `Set`?

atzedus:
> Why not just pass multiple `SetCol` in place of `Set`?

For fixed values, `SetCol`/`SetExpr` are definitely the better choice. But here we're specifically talking about `Set(...)`, i.e., when the list of assignments is built dynamically, like:

```go
sets := []bob.Expression{
  psql.Quote("users", "name").OP("=", psql.Arg("Bob")),
  psql.Quote("users", "surname").OP("=", psql.Arg("Bob")),
}

if withLastName {
  sets = append(sets, psql.Quote("users", "name").OP("=", psql.Arg("Bob")))
}

um.Set(sets...)
```

That said, I think you're right overall — introducing `Assign` is probably overengineering.

atzedus:
@stephenafamo EQ now behaves differently for SET and similar constructs. PR ready.

stephenafamo:
Much better. I like this.

I assume the performance of `SET` would take a hit, but probably a fair tradeoff.

Thanks once again

PR #702, opened 2026-06-03
fix(psql): preserve standalone joins before FROM/USING and support join chains in sm.From

atzedus:
Fixes PostgreSQL query builder behavior when joins are applied **before** the primary `FROM` / `USING` source, and extends `sm.From` to accept inline join chains.

- **UPDATE (`um`) / DELETE (`dm`)**: Standalone join mods (`InnerJoin`, `LeftJoin`, etc.) applied before `um.From(...)` or `dm.Using(...)` are no longer dropped. `AppendJoin` now creates a placeholder `from_item` when none exists yet; `AppendTableRef` fills that slot and merges any accumulated joins into the real table ref.
- **SELECT (`sm`)**: `sm.From(table, joins...)` now accepts optional join chains (aligned with `um.From` / `dm.Using`). Inline joins passed to `From` are merged with joins already attached via standalone mods (e.g. `CrossJoin` then `From(..., InnerJoin(...))`).

## Problem

Previously, patterns like:

```go
psql.Select(
    sm.InnerJoin("events").OnEQ(...),
    sm.From("users"),
)
```

or UPDATE/DELETE with a join mod before `From`/`Using` silently omitted the join. `sm.From` also could not attach joins inline in a single mod.

## Solution

- `UpdateQuery` / `DeleteQuery`: mirror the join-placeholder + merge-on-`AppendTableRef` pattern already used for `SelectQuery`.
- `sm.From`: variadic `JoinChain` args, same as other `From` helpers.
- Regression tests for SELECT, UPDATE, and DELETE covering join-before-from/using and inline/merged join chains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `NameAsExpr()` on dialect `View` / `Table` types (PostgreSQL, SQLite, MySQL) omits a redundant `AS` when the alias matches the table name (for example `FROM "users"` instead of `FROM "users" AS "users"`). An explicit alias is still emitted when a schema is set at construction (`AS "schema.table"`). (thanks @atzedus)
 
+### Fixed
+
+- Fixed PostgreSQL `um` / `dm` standalone join mods (`InnerJoin`, `LeftJoin`, etc.) applied before `um.From(...)` / `dm.Using(...)`: the join is now preserved and attached to the next `FROM` / `USING` from_item instead of being dropped. (thanks @atzedus)
+- Fixed/extended PostgreSQL `sm.From(...)`: it now accepts variadic join chains (`sm.From(table, joins...)`), and inline joins are merged with already attached standalone joins when replacing the primary `FROM` source. (thanks @atzedus)
+
 ## [v0.45.0] - 2026-05-28
 
 ### Added
--- a/dialect/psql/delete_test.go
+++ b/dialect/psql/delete_test.go
@@ -75,6 +75,21 @@ func TestDelete(t *testing.T) {
 			  WHERE (employees.id = $1)`,
 			ExpectedArgs: []any{1},
 		},
+		"standalone inner join before using": {
+			Query: psql.Delete(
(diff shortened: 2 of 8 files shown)


atzedus:
Hi @stephenafamo,

Following up on the discussion in #696 (specifically [this comment](https://github.com/stephenafamo/bob/pull/696#issuecomment-4543226803)), I have analyzed the current mechanisms in `bob` and concluded that the direction proposed there is actually incorrect. 

In this PR, I have only added `(...joins)` to the `sm.From` helper to synchronize it with `um.From` and `dm.From`/`Using`. However, I strongly believe that the idea of removing the separate `InnerJoin`, `CrossJoin`, and other join helpers is a mistake. Here is why:

### 1. The "Add vs. Replace" Philosophy of Mods
The core concept in `bob` is that mods which *add* to a query always append, while mods that *set* a value replace the previously added one. The current `sm/um/dm.From` and `Using` mechanisms work exactly as "setters/replacers". 

### 2. Loss of Flexibility (The `sql-to-code` use case)
If we remove the separate `InnerJoin`/`CrossJoin` mods for `sm`, we lose the flexibility to modify queries. The most common and powerful scenario for this is `sql-to-code` generation. 
In this workflow, we generate a "base" SELECT query, and then use specific mods like `Columns`, `InnerJoin`, and `CrossJoin` to safely append necessary fields and join tables. If we remove these helpers, it becomes impossible to do this safely. The only alternative would be using `sm.From`, but since `From` acts as a replacer, it would overwrite the base table and wipe out all previously added joins.

### 3. Consistency in `um` and `dm`
Regarding the specific changes made in this PR for `um.From`/`Using` and `dm.From`/`Using`: in the vast majority of cases, these are used with a single target table. I consider it incorrect to silently ignore an `InnerJoin` that was applied *before* `From`. In `sm.From`, this works differently (it handles them properly), so I updated `um` and `dm` to ensure consistent behavior across all mods.

atzedus:
@stephenafamo Let me know if you have any questions or if any adjustments are required.

atzedus:
@stephenafamo Could you please make a patch release 0.45.1 with the current fixes?

PR #703, opened 2026-06-04
fix: rollback transaction on panic in RunInTx

masashi-toda-ene:
## Problem

`RunInTx` currently does not handle panics inside the provided function.
If `fn` panics, the transaction is never rolled back, leaving row-level
locks held until the connection is GC'd or the server's
`innodb_lock_wait_timeout` fires — which can cause deadlocks under
concurrent load.

```go
db.RunInTx(ctx, nil, func(ctx context.Context, exec bob.Executor) error {
    // INSERT acquires row locks...
    panic("something went wrong") // locks are never released
})
```

## Fix

Add a deferred `recover()` that rolls back the transaction and re-panics,
matching the behaviour of other Go ORMs (GORM, go-pg, gobuffalo/pop).

- If rollback succeeds: re-panic with the original value unchanged
- If rollback fails **and** the panic value is an `error`: join both errors via `errors.Join` (consistent with the existing error-path style) and re-panic
- If rollback fails **and** the panic value is not an `error` (e.g. a string): rollback error is intentionally discarded — changing the panic value type would break callers who `recover()` and type-assert the value

## Tests

Three new tests added to `stdlib_test.go` using a minimal mock `driver.Connector`:

| Test | Verifies |
|---|---|
| `TestRunInTxPanicPropagates` | panic re-raised after successful rollback |
| `TestRunInTxPanicErrorJoinsRollbackErr` | both errors visible via `errors.Is` when rollback also fails |
| `TestRunInTxPanicNonErrorPreservesValue` | original non-error panic value preserved |

## References

- go-pg panic recovery: https://github.com/go-pg/pg/issues/468
- gobuffalo/pop: https://github.com/gobuffalo/pop/pull/775

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -64,11 +64,26 @@ func (d DB) BeginTx(ctx context.Context, opts *sql.TxOptions) (Tx, error) {
 // RunInTx runs the provided function in a transaction.
 // If the function returns an error, the transaction is rolled back.
 // Otherwise, the transaction is committed.
+// If the function panics, the transaction is rolled back and the panic is re-raised.
 func (d DB) RunInTx(ctx context.Context, txOptions *sql.TxOptions, fn func(context.Context, Executor) error) error {
 	tx, err := d.BeginTx(ctx, txOptions)
 	if err != nil {
 		return fmt.Errorf("begin: %w", err)
 	}
+
+	defer func() {
+		if p := recover(); p != nil {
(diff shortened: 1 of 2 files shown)


stephenafamo:
Kindly fix the lint errors and merge conflicts.
Also add a note in CHANGELOG.md along with (thanks @yourhandle)

PR #704, opened 2026-06-04
feat: augment generated queries

daddz:
Adds a new `With()` function to `ModQuery` and `ExecModQuery` to conveniently add additional mods to generated queries.

### DIFF
--- a/website/docs/code-generation/queries.md
+++ b/website/docs/code-generation/queries.md
@@ -153,6 +153,24 @@ query := sqlite.Select(
 )
 ```
 
+#### Augmenting with `With()`
+
+As a shortcut, the generated query also has a `With()` method that applies extra mods on top of the generated query and returns a runnable query, so you can keep the finishers without re-wrapping it in `Select`:
+
+```go
+// Also filter where name = "Bob", and only return 10 rows
+users, err := AllUsers(1).With(
+    sm.Where(psql.Quote("name").EQ(psql.Arg("Bob"))),
+    sm.Limit(10),
+).All(ctx, db)
+```
+
+:::caution
+
+`With()` (like using the query as a mod) only **appends** clauses. For a `WHERE` clause this is fine: the extra condition is joined to the existing one with `AND`. But if the generated query already has an `ORDER BY`, `LIMIT`, or `OFFSET` clause, adding another one does **not** replace it — both are emitted, producing invalid SQL. Only add these clauses when the base query does not already define them.
+
+:::
+
 ## Annotating queries
 
 Each query has the following attributes that can be modified with annotations:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `With()` on generated queries as a more ergonomic method to add mods on top of generated queries (thanks @daddz)
+
 ### Changed
 
 - `EQ` comparisons passed to `SET` / `ON CONFLICT DO UPDATE SET` / `MERGE ... UPDATE SET` / MySQL `ON DUPLICATE KEY UPDATE` render without the extra parentheses used in `WHERE` / `ON` (e.g. `SET "users"."name" = $1` vs `WHERE ("users"."name" = $1)`). (thanks @atzedus)
-
 - `NameAsExpr()` on dialect `View` / `Table` types (PostgreSQL, SQLite, MySQL) omits a redundant `AS` when the alias matches the table name (for example `FROM "users"` instead of `FROM "users" AS "users"`). An explicit alias is still emitted when a schema is set at construction (`AS "schema.table"`). (thanks @atzedus)
 
 ## [v0.45.0] - 2026-05-28
--- a/gen/templates/queries/query/01_query.go.tpl
+++ b/gen/templates/queries/query/01_query.go.tpl
@@ -6,6 +6,7 @@
 {{$.Importer.Import "context"}}
 {{$.Importer.Import "github.com/stephenafamo/bob"}}
 {{$.Importer.Import "github.com/stephenafamo/bob/orm"}}
+{{$.Importer.Import (printf "github.com/stephenafamo/bob/dialect/%s" $.Dialect)}}
 {{$.Importer.Import (printf "github.com/stephenafamo/bob/dialect/%s/dialect" $.Dialect)}}
 
 //go:embed {{.QueryFile.BaseName}}.bob.sql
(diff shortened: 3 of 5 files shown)


website/docs/code-generation/queries.md
```diff
@@ -153,6 +153,24 @@ query := sqlite.Select(
...
+    sm.Where(psql.Quote("name").EQ(psql.Arg("Bob"))),
+    sm.Limit(10),
+).All(ctx, db)
+```
+
+:::caution
+
+`With()` (like using the query as a mod) only **appends** clauses. For a `WHERE` clause this is fine: the extra condition is joined to the existing one with `AND`. But if the generated query already has an `ORDER BY`, `LIMIT`, or `OFFSET` clause, adding another one does **not** replace it — both are emitted, producing invalid SQL. Only add these clauses when the base query does not already define them.
+
+:::
```
stephenafamo, line 172: Can you test this for sure... this shouldn't work like this.... adding these should not cause an invalid query.

* Adding `ORDER BY` should append a new ordering term
* Adding `LIMIT` or `OFFSET` should overwrite it.

stephenafamo (changes requested):

stephenafamo:
Also please fix the merge conflicts

website/docs/code-generation/queries.md
daddz: There are tests for this behaviour: https://github.com/stephenafamo/bob/pull/704/changes#diff-5d1c75af7bb52207477503aa064209b5bf906cfb277ac7ffaa38490adedf6885R104

The same thing happens currently with 0.45:
```sql
-- ListTodosWithOrderAndLimit
SELECT * FROM todo WHERE user_id = $1 ORDER BY id DESC LIMIT 2 OFFSET 1;
```

```go
query := psql.Select(queries.ListTodosWithOrderAndLimit(1), sm.OrderBy(psql.Quote("title")), sm.Limit(2), sm.Offset(1))
todos, err := bob.All(ctx, dbDebug, query, scan.StructMapper[models.Todo]())
```
leads to:
```sql
SELECT 
"todo"."id" AS "id", "todo"."title" AS "title", "todo"."description" AS "description", "todo"."is_completed" AS "is_completed", "todo"."project_id" AS "project_id", "todo"."user_id" AS "user_id", "todo"."due_date" AS "due_date", "todo"."created_at" AS "created_at"
FROM todo
WHERE user_id = $1
ORDER BY "title"
LIMIT 2
OFFSET 1
ORDER BY id DESC
LIMIT 2
OFFSET 1
```
```
 pq: syntax error at or near "ORDER"
```

daddz:
> Also please fix the merge conflicts

Fixed, suddenly the tests are failing but I don't think it has anything to do with my changes and might be a fluke with sqlite.

website/docs/code-generation/queries.md
stephenafamo: I'll need to find time to understand why this is happening before merging. I feel like I'm missing some detail 😫

daddz: From what I understand it's related to the `Combined...` fields that are generated for psql e.g.:
```sql
-- ListTodosWithOrderLimitUnion
(SELECT * FROM todo WHERE user_id = $1 ORDER BY id DESC LIMIT 2 OFFSET 1)
UNION
(SELECT * FROM todo WHERE user_id = $1 ORDER BY id DESC LIMIT 2 OFFSET 1)
ORDER BY id DESC LIMIT 2 OFFSET 1;
```
```go
Mod: bob.ModFunc[*dialect.SelectQuery](func(q *dialect.SelectQuery) {
	q.AppendSelect(expressionTypArgs.subExpr(8, 272))
	q.SetTable(expressionTypArgs.subExpr(278, 282))
	q.AppendWhere(expressionTypArgs.subExpr(289, 301))
	q.CombinedOffset.SetOffset(expressionTypArgs.subExpr(334, 335))

	q.AppendCombine(clause.Combine{
		Strategy: "UNION",
		All:      false,
		Query: bob.BaseQuery[bob.Expression]{
			Expression: expressionTypArgs.subExpr(344, 415),
			QueryType:  bob.QueryTypeSelect,
			Dialect:    dialect.Dialect,
		},
	})
	q.CombinedOrder.AppendOrder(expressionTypArgs.subExpr(426, 433))
	q.CombinedLimit.SetLimit(expressionTypArgs.subExpr(440, 441))
}),
```

And calling it with additional mods:
```go
query := psql.Select(queries.ListTodosWithOrderLimitUnion(1), sm.OrderBy(psql.Quote("title")), sm.Limit(2), sm.Offset(1))
result, err := bob.All(ctx, dbDebug, query, scan.StructMapper[models.Todo]())
```
```sql
(SELECT 
"todo"."id" AS "id", "todo"."title" AS "title", "todo"."description" AS "description", "todo"."is_completed" AS "is_completed", "todo"."project_id" AS "project_id", "todo"."user_id" AS "user_id", "todo"."due_date" AS "due_date", "todo"."created_at" AS "created_at"
FROM todo
WHERE user_id = $1
ORDER BY "title"
LIMIT 2
OFFSET 1)
UNION (SELECT * FROM todo WHERE user_id = $2 ORDER BY id DESC LIMIT 2 OFFSET 1)
ORDER BY id DESC
LIMIT 2
OFFSET 1
```
As you can see it applies the mods to the first select query instead of the outer-most.

daddz: The following mods work though, not sure if that was the actual expectation to use these:
```go
queries.ListTodosWithOrderLimitUnion(1), sm.OrderCombined(psql.Quote("title")), sm.LimitCombined(5), sm.OffsetCombined(10)
```
```sql
SELECT 
"todo"."id" AS "id", "todo"."title" AS "title", "todo"."description" AS "description", "todo"."is_completed" AS "is_completed", "todo"."project_id" AS "project_id", "todo"."user_id" AS "user_id", "todo"."due_date" AS "due_date", "todo"."created_at" AS "created_at"
FROM todo
WHERE user_id = $1
UNION (SELECT * FROM todo WHERE user_id = $2 ORDER BY id DESC LIMIT 2 OFFSET 1)
ORDER BY id DESC, "title"
LIMIT 5
OFFSET 10
```

But it also swallows the existing ORDER/LIMIT/OFFSET clauses from the first query.

stephenafamo: Yes, I think the expectation is to use the "Combined"

Alternatively, the parser could be tweaked to not use combined if there is no union

daddz: > Yes, I think the expectation is to use the "Combined"

Okay so that means it works as expected right now and with this PR, right?

> Alternatively, the parser could be tweaked to not use combined if there is no union

I'd take a stab at that in a follow-up PR and also extend the documentation around this behaviour in psql - or would you prefer that all-in-one?

daddz: @stephenafamo: I split out the parser fix so that may be merged first https://github.com/stephenafamo/bob/pull/710

stephenafamo:
Kindly fix the merge conflicts 🙏🏾

stephenafamo (approved):

PR #706, opened 2026-06-07
perf(gen): stitch slice relationship loaders in O(N+M) via a map

sandonemaki:
## Summary

Generated slice relationship loaders (`<Parent>Slice.Load<Relationship>`) stitch the
loaded parents and children with an **O(N×M) nested loop**. For relationships with many
rows on both sides this dominates load time. This PR stitches them in **O(N+M)** using a
map keyed by the join column, while keeping the existing behaviour.

## Motivation

In a real eager/then-load of a relationship with many parents and children, the
non-query Go time was:

| | non-query Go time |
|---|---|
| before | ~178 s |
| after | ~0.8 s |

The query itself was already fast; the cost was the `O(N×M)` stitching loop.

## What changed

`gen/templates/loaders/table/110_loaders.go.tpl` — both slice loaders
(direct join, and through a join table) now build a map `joinKey -> []*Parent`
and stitch by looking each child up once:

```go
// Example: UserSlice.LoadPosts (a User has many Posts: post.user_id = user.id)

// before: O(N×M)
for _, o := range os {
    for _, rel := range posts {
        if !(o.ID == rel.UserID) { continue }
        rel.R.User = o
        o.R.Posts = append(o.R.Posts, rel)
    }
}

// after: O(N+M)
userByKey := make(map[int64][]*User, len(os))
for _, o := range os {
    if o == nil { continue }
    userByKey[o.ID] = append(userByKey[o.ID], o)
}
for _, rel := range posts {
    owners, ok := userByKey[rel.UserID]
    if !ok { continue }
    for _, o := range owners {
        rel.R.User = o
        o.R.Posts = append(o.R.Posts, rel)
    }
}
```

Two small helpers were added to `gen/drivers/types.go`:

- `CanCompareWithEquals` — whether a type is compared with `==` (no custom `compare_expr`),
  i.e. usable as a Go map key.
- `UnwrapNullExpr` — extracts the underlying value of a nullable column for use as a map key.

## Safety / behaviour preservation

The map path is taken **only when it is provably equivalent**, otherwise it falls back to
the original nested loop, so generated behaviour is unchanged:

- **Single-column, `==`-comparable joins only.** Composite keys and types with a custom
  `compare_expr` (e.g. `[]byte`, which cannot be a Go map key) fall back to the nested loop.
- **NULL handling.** Nullable join columns are keyed by their underlying value and NULL rows
  are skipped (NULL never matches in SQL), matching the previous `continue` guards.
- **to-one.** The map version keeps only the first matching child
  (`if o.R.<rel> != nil { continue }`), matching the previous `break`.
- The same parent key may map to several parents, so the value is `[]*Parent`.

## Testing

- `golangci-lint run` → `0 issues.`
- `gofumpt -l` → no diff
- `go test ./gen/bobgen-sqlite/driver/` → generation + build + generated tests pass on the
  pure-Go `modernc` and `ncruces` drivers (covering direct joins, many-to-many through a
  join table, nullable FKs, and to-one / to-many).

## Checklist

- [x] `gofumpt -l` reports no changes
- [x] generated code compiles and the generated tests pass (sqlite, modernc/ncruces)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `EQ` comparisons passed to `SET` / `ON CONFLICT DO UPDATE SET` / `MERGE ... UPDATE SET` / MySQL `ON DUPLICATE KEY UPDATE` render without the extra parentheses used in `WHERE` / `ON` (e.g. `SET "users"."name" = $1` vs `WHERE ("users"."name" = $1)`). (thanks @atzedus)
 - `NameAsExpr()` on dialect `View` / `Table` types (PostgreSQL, SQLite, MySQL) omits a redundant `AS` when the alias matches the table name (for example `FROM "users"` instead of `FROM "users" AS "users"`). An explicit alias is still emitted when a schema is set at construction (`AS "schema.table"`). (thanks @atzedus)
 
+- Generated slice relationship loaders (`<Parent>Slice.Load<Relationship>`) now stitch loaded parents and children in O(N+M) using a map keyed by the join column, instead of the previous O(N×M) nested loop. This greatly speeds up eager-loading and then-loading relationships with many rows on both sides. The map is only used for single-column joins whose key type is comparable with `==`; composite keys and types with a custom `compare_expr` (e.g. `[]byte`) fall back to the original nested loop, so generated behaviour is unchanged. (thanks @sandonemaki)
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


stephenafamo:
Kindly Update CHANGELOG.md

sandonemaki:
@stephenafamo 
Updated CHANGELOG.md as requested.

Also added a small follow-up, fix(gen): skip nil parents in via-join slice relationship loaders — the multi-side loader's reset loop and nested-loop fallback were missing the nil check that the other loader paths already have.

This failure (sqlite3: database is locked in generate_with_aliases) looks like a pre-existing flaky test caused by concurrent SQLite access in CI. Could you re-run the failing job? The test passes locally.

sandonemaki:
@stephenafamo 
Regarding the CI failures: the failing test changes every run, which confirms this is a race condition in the test suite rather than an issue with my changes.

Run 1: ncruces/generate_with_aliases
Run 2: modernc/generate
Run 3: ncruces/generate

All fail with sqlite3: database is locked (SQLITE_BUSY) — multiple SQLite test configurations running concurrently are competing for the same database.

PR #708, opened 2026-06-09
feat(codegen): add column_order config option

cbarber:
## Problem

When multiple developers apply migrations independently, the `ordinal_position` of columns diverges between their local databases. Since bob generates models ordered by `ordinal_position`, two developers generating from the same logical schema produce different files — causing spurious diffs in source control.

## Solution

Add a `column_order` config option (shared across all three drivers via `helpers.Config`) with two values:

- `"ordinal"` (default) — preserves previous behaviour, no breaking change
- `"name"` — sorts columns alphabetically after retrieval, producing deterministic output regardless of migration order

The sort is applied post-retrieval in `drivers.BuildDBInfo` (psql, mysql) and `driver.getTable` (sqlite), keeping the change minimal and consistent.

## Usage

```yaml
# bobgen.yaml
column_order: name
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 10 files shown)


stephenafamo:
Update the documentation and CHANGELOG.md

cbarber:
Updated. Do you want the markdown table pipes to be aligned or just let the markdown processor deal with it?

stephenafamo:
> Do you want the markdown table pipes to be aligned or just let the markdown processor deal with it?

Yes please. Not a big deal though

PR #709, opened 2026-06-10
fix: index increment in debug output

jay-babu:
before, args would be printed 
0
1
2

while the sql would have $1, $2. just matching the numbers now


stephenafamo:
I'm not sure about this, I know SQL is mostly a 1-based index, but it still feel wrong, unless we provide a way for dialects to report their starting index (maybe an optional interface).

Alternatively, you can create your own debug executor.

jay-babu:
forgot about other dbs. thanks!

PR #710, opened 2026-06-10
fix(psql): fix generated query mods for combined queries (UNION/INTERSECT/EXCEPT)

daddz:
The psql query parser was generating incorrect mods for combined queries. It always routed the top-level `ORDER BY/LIMIT/OFFSET/FETCH` into the `Combined...` clauses, and never handled the first queries' own clauses. 

A query like:
```sql
-- MyQuery
(SELECT ... ORDER BY id DESC LIMIT 2 OFFSET 1)
UNION
(SELECT ...)
ORDER BY id DESC LIMIT 5 OFFSET 6
```
```go
query := psql.Select(queries.MyQuery(), sm.LimitCombined(10), sm.OffsetCombined(10), sm.OrderCombined("id"))
```
would lose the first queries' `ORDER BY/LIMIT`, misroute its `OFFSET` onto the combined result, and drop the outer OFFSET, so the generated query no longer matched the SQL it was parsed from.

This change splits the handling to mirror how `dialect.SelectQuery` actually renders. 
The first queries' `ORDER BY/LIMIT/OFFSET/FETCH` map to the regular query clauses, while the outer-most clauses map to `CombinedOrder/CombinedLimit/CombinedOffset/CombinedFetch` (only when combines are present).


stephenafamo:
Thank you 🙏🏾

PR #714, opened 2026-06-13
perf(gen): use col = ANY(\$1) for single-column relationship loaders on psql

sandonemaki:
## Summary

For the PostgreSQL dialect, two generated `<Parent>Slice` helpers filter children by the
parent-PK set with `fk IN (SELECT unnest(CAST($1 AS T[])))`:

- the relationship query `func (os <Parent>Slice) <Relationship>(...)` (used by the slice
  eager / then loaders), and
- the batch count method `func (os <Parent>Slice) LoadCount<Rel>(...)` (counts a to-many
  relationship for many parents in one `GROUP BY` query).

This PR changes the **single-column** case of both to `fk = ANY(CAST($1 AS T[]))`, so the
PostgreSQL planner can use an index scan on the foreign key instead of a
`Seq Scan + Hash Semi Join`. Composite keys and other dialects are unchanged.

## Generated code & SQL — before / after

Example: `users` (parent) → `posts` (child, `posts.user_id -> users.id`, `bigint`),
i.e. `UserSlice.Posts(...)`.

**Before**

```go
PKArgExpr := psql.Select(sm.Columns(
    psql.F("unnest", psql.Cast(psql.Arg(pkID), "bigint[]")),
))
sm.Where(psql.Group(Posts.Columns.UserID).OP("IN", PKArgExpr)),
```

```sql
WHERE posts.user_id IN (SELECT unnest(CAST($1 AS bigint[])))
```

**After**

```go
PKArgExpr := psql.Any(psql.Cast(psql.Arg(pkID), "bigint[]"))
sm.Where(Posts.Columns.UserID.EQ(PKArgExpr)),
```

```sql
WHERE posts.user_id = ANY(CAST($1 AS bigint[]))
```

## Query plan

**Before — `fk IN (SELECT unnest($1))`**

```text
Hash Semi Join
  Hash Cond: (posts.user_id = (unnest($1)))
  ->  Seq Scan on posts                 -- full table scan (the slowdown)
  ->  Hash
        ->  Function Scan on unnest      -- fixed row estimate; $1 length unknown at plan time
```

**After — `fk = ANY($1)`**

```text
Bitmap Heap Scan on posts
  Recheck Cond: (user_id = ANY ($1))
  ->  Bitmap Index Scan on posts_user_id_idx
        Index Cond: (user_id = ANY ($1))  -- the FK index is used
```

## Why

| | Before (`Seq Scan` + `Hash Semi Join`) | After (`Bitmap Index Scan`) |
| --- | --- | --- |
| Generated SQL | `user_id IN (SELECT unnest(CAST($1 AS bigint[])))` | `user_id = ANY(CAST($1 AS bigint[]))` |
| FK index | not used (absent from the plan) | used (`posts_user_id_idx`) |
| Rows read | all of `posts` (every heap page) | only matching rows |
| Cost | scales with table size (O(rows)) | scales with the result size |
| Ops | needs `pg_hint_plan` hints | no hints, stable |
| Reason | `IN (subquery)` is planned as a semi-join; `unnest()` has a fixed row estimate and the `$1` array length is unknown at plan time, so the planner falls back to scanning the whole table and the index cannot be applied | `= ANY(array)` is a `ScalarArrayOpExpr`, not a join, so the planner takes the FK index path directly |

## Scope

- Two templates, same change: `gen/templates/models/table/009_rel_query.go.tpl` (slice
  relationship query) and `gen/templates/counts/table/115_counts.go.tpl` (batch count
  method). The count method's example is identical — the same 2-line edit
  (`= ANY` + `.EQ(...)`) on the SELECT it builds.
- psql + single join column only. Composite keys keep `(c1, c2) IN (SELECT unnest..., unnest...)`
  (a row tuple cannot use `= ANY`); other dialects are unchanged.

## EXPLAIN ANALYZE benchmark (PostgreSQL 18.4, index on FK column)

Tested with PREPARE/EXECUTE (parametrized `$1`, matching actual generated code) against
a child table where each parent has ~1,000 child rows — a typical one-to-many workload.
The **Returned rows** column confirms both forms return identical results.

| Rows | Batch | Before time | Before plan | After time | After plan | Speedup | Returned rows |
|---|---|---|---|---|---|---|---|
| 100K | 5 | 6.1 ms | Seq Scan + Hash Semi Join | 0.77 ms | Bitmap Index Scan | **7.9×** | 4,559 |
| 100K | 50 | 9.7 ms | Seq Scan + Hash Semi Join | 4.0 ms | Bitmap Index Scan | **2.4×** | 50,035 |
| 1M | 5 | 3.1 ms | Nested Loop (5× Bitmap Index) | 1.4 ms | Bitmap Index Scan | **2.2×** | 4,513 |
| 1M | 50 | 35.5 ms | Seq Scan + Hash Semi Join | 10.2 ms | Bitmap Index Scan | **3.5×** | 49,459 |
| 10M | 5 | 19.1 ms | Nested Loop (5× Bitmap Index) | 2.4 ms | Bitmap Index Scan | **8.0×** | 4,552 |
| 10M | 50 | 115.6 ms | Nested Loop (50× Bitmap Index) | 41.4 ms | Bitmap Index Scan | **2.8×** | 49,592 |
| 10M | 200 | 278 ms | Seq Scan + Hash Join | 79.8 ms | Bitmap Index Scan | **3.5×** | 199,316 |
| 50M | 5 | 134 ms | Nested Loop (5× Bitmap Index) | 7.8 ms | Bitmap Index Scan | **17×** | 4,539 |
| 50M | 50 | 1,087 ms | Nested Loop (50× Bitmap Index) | 171 ms | Bitmap Index Scan | **6.4×** | 49,530 |
| 50M | 200 | 2,508 ms | Nested Loop (200× Bitmap Index) | 504 ms | Bitmap Index Scan | **5.0×** | 200,508 |

### Production database validation (PostgreSQL 18, warm cache)

The same SQL forms were run against a production application database to confirm
the synthetic results hold on real data. The FK column has a B-tree index in both cases.

**Case 1 — simple FK lookup (batch = 20)**

| | Before (`IN unnest`) | After (`= ANY`) |
|---|---|---|
| Execution time | 192 ms | 12 ms |
| Scan on child table | Index Scan × 20 (one per FK value) | Bitmap Index Scan × 1 |
| Speedup | | **16×** |

The planner consolidated 20 individual index seeks into a single Bitmap Index Scan,
matching the pattern shown in the synthetic benchmark.

**Case 2 — FK lookup with JOIN to a large related table (batch = 1)**

| | Before (`IN unnest`) | After (`= ANY`) |
|---|---|---|
| Execution time | 901 ms | 11 ms |
| Join strategy | Merge Join — Index Only Scan on related table (1.4 M rows) | Nested Loop — 73 × Index Scan (~10 rows each) |
| Speedup | | **82×** |

`= ANY` gave the planner enough information to switch from Merge Join (which required
a full ordered scan of 1.4 M rows in a related table) to Nested Loop with targeted
index lookups, reducing rows read from ~1.4 M to ~730. This shows the improvement
reaches beyond the child table itself: accurate FK estimation affects join strategy
for the entire query.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed generated `Setter.Apply` wrapping all insert values in a single `ExpressionFunc`, which prevented `BeforeInsertHooks` and query mods from modifying individual column values via `q.Values.Vals[row][columnIndex]`. Base and SQLite codegen templates now emit one expression per column (MySQL already did). Generated SQL is unchanged. (thanks @atzedus)
+- Fixed generated query mods for combined `UNION/INTERSECT/EXCEPT` queries dropping or misrouting the first operand's `ORDER BY/LIMIT/OFFSET` into the combined-result clauses. (thanks @daddz)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed PostgreSQL `um` / `dm` standalone join mods (`InnerJoin`, `LeftJoin`, etc.) applied before `um.From(...)` / `dm.Using(...)`: the join is now preserved and attached to the next `FROM` / `USING` from_item instead of being dropped. (thanks @atzedus)
 - Fixed/extended PostgreSQL `sm.From(...)`: it now accepts variadic join chains (`sm.From(table, joins...)`), and inline joins are merged with already attached standalone joins when replacing the primary `FROM` source. (thanks @atzedus)
 
+### Fixed
+
+- Fixed generated `Setter.Apply` wrapping all insert values in a single `ExpressionFunc`, which prevented `BeforeInsertHooks` and query mods from modifying individual column values via `q.Values.Vals[row][columnIndex]`. Base and SQLite codegen templates now emit one expression per column (MySQL already did). Generated SQL is unchanged. (thanks @atzedus)
+
 ### Changed
 
 - Moved PostgreSQL `SetCols` tuple-assignment builder from `dialect/psql/dialect/setcols.go` to shared `clause/setcols.go` (`clause.NewSetCols`, `clause.SetColsOptions`). Dialect wrappers configure `ToRow` via `SetColsOptions.RowPrefix` (e.g. `"ROW"` on PostgreSQL). (thanks @atzedus)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- PostgreSQL single-column relationship loaders — both the slice relationship query (`<Parent>Slice.<Rel>`) and the batch count method (`<Parent>Slice.LoadCount<Rel>`) — now filter related rows with `col = ANY($1)` instead of `col IN (SELECT unnest($1))`, so the query planner can use an index scan on the foreign key instead of a `Seq Scan + Hash Semi Join`. Composite-key relationships keep the existing `IN (SELECT unnest(...))` form. (thanks @sandonemaki)
+
 ## [v0.46.0] - 2026-06-11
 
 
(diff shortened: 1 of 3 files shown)


stephenafamo:
Thank you for this. Only a few changes needed:

1. No need for the comments in the template. Or at least maybe make it shorter keeping only the English comment)
1. Update CHANGELOG.md

CHANGELOG.md
```diff
@@ -26,6 +27,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Generated slice relationship loaders (`<Parent>Slice.Load<Relationship>`) now stitch loaded parents and children in O(N+M) using a map keyed by the join column, instead of the previous O(N×M) nested loop. This greatly speeds up eager-loading and then-loading relationships with many rows on both sides. The map is only used for single-column joins whose key type is comparable with `==`; composite keys and types with a custom `compare_expr` (e.g. `[]byte`) fall back to the original nested loop, so generated behaviour is unchanged. (thanks @sandonemaki)
+- PostgreSQL single-column relationship loaders — both the slice relationship query (`<Parent>Slice.<Rel>`) and the batch count method (`<Parent>Slice.LoadCount<Rel>`) — now filter related rows with `col = ANY($1)` instead of `col IN (SELECT unnest($1))`, so the query planner can use an index scan on the foreign key instead of a `Seq Scan + Hash Semi Join`. Composite-key relationships keep the existing `IN (SELECT unnest(...))` form. (thanks @sandonemaki)
```
stephenafamo: This should be under a new heading of `[Unreleased]`

CHANGELOG.md
```diff
@@ -44,7 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **BREAKING:** MySQL `sm.ForUpdate` and `sm.ForShare` now take `tables ...any` with the same semantics as PostgreSQL (`mysql.Quote(...)` when needed). (thanks @atzedus)
 - **BREAKING:** `clause.Lock.Tables` is now `[]any` instead of `[]string` (used by `FOR UPDATE` / `FOR SHARE` rendering). (thanks @atzedus)
 - **BREAKING:** `View` / `Table` `Name()` now returns the bare table (or view) name as `string`. In v0.44.0, `Name()` returned a dialect `Expression` (qualified table reference for SQL). That role moved to new methods:
-  - `NameExpr()` — what `Name()` did in v0.44.0 (on PostgreSQL and SQLite, respects `UseSchema` when schema was generated empty)
+  - `NameExpr()` — what `Name()` did in v0.44.0 (on PostgreSQL and SQLite, respects `UseSchema` whens schema was generated empty)
```
stephenafamo: Mistake?

stephenafamo (changes requested):

CHANGELOG.md
sandonemaki: @stephenafamo 
I've applied the fix. However, CI timed out. Locally, the test took ~54s without -race, so the CI timeout limit may need to be increased.

sandonemaki:
@stephenafamo 
Tests are passing locally with the following times:
ok      github.com/stephenafamo/bob/gen/bobgen-sqlite/driver    16.735s
ok      github.com/stephenafamo/bob/gen/bobgen-psql/driver      53.992s

sandonemaki:
@stephenafamo 
The PR description includes a production database validation section.

Beyond the synthetic benchmark, I ran `EXPLAIN ANALYZE` against a real application database (PostgreSQL 18) and confirmed the improvement holds on production data — **16× speedup** on a simple FK lookup (batch 20), and **82× on a query with a JOIN** to a large related table, where `= ANY` caused the planner to switch from Merge Join (1.4 M rows scanned) to Nested Loop (~730 rows).

PR #717, opened 2026-06-18
feat(dberrors): add PostgreSQL check constraint errors

keithbro-imx:
## Summary

- Generate `CheckConstraintError` and per-table `ErrCheck...` constants from check constraints in the `dberrors` package
- Match PostgreSQL check violations (SQLSTATE `23514`) by constraint name for `pq` and `pgx` drivers
- Add synthetic generated tests and update code generation docs

## Why

Bob already generates typed errors for unique constraint violations, but check constraint violations were not covered. On PostgreSQL these surface as SQLSTATE `23514`, so callers had to match raw driver errors instead of using the same `errors.Is` pattern as unique constraints.

## Before / After

Before, the `dberrors` plugin only generated `UniqueConstraintError` constants. Check constraint violations from PostgreSQL had to be handled by inspecting raw `pq.Error` / `pgconn.PgError` values.

After this change, tables with check constraints get generated `ErrCheck{Name}` constants plus a generic `ErrCheckConstraint`, and PostgreSQL drivers match `23514` by constraint name. Unique constraint error generation and matching are unchanged.

## Test plan

- [x] `go test ./gen/bobgen-psql/driver`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Generated `dberrors` packages now include generic and per-table check-constraint errors for PostgreSQL, matched by constraint name for `pq` and `pgx` drivers. (thanks @keithbro-imx)
+
 ## [v0.46.0] - 2026-06-11
 
 ### Added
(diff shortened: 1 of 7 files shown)


keithbro-imx:
Hi @stephenafamo - thanks for creating such a good library. First time contributor here - noticed that we generate code for unique constraints but not check constraints. Let me know if you have any questions on the PR.

stephenafamo:
Add a note in CHANGELOG.md under `[Unreleased]` with a note of your changes (and thanks)

stephenafamo:
Kindly Fix conflicts

keithbro-imx:
Thank you @stephenafamo

PR #720, opened 2026-07-01
Fix column order on sql query star type mismatch

cbarber:
## Problem

When `column_order: "name"` is set, the psql driver sorts table columns
alphabetically in `w.db` before query parsing. Custom SQL queries using
`.*` (e.g. `SELECT u.* FROM users AS u`) expand the wildcard using that
sorted column list, producing `source.columns` in alphabetical order.

However, `getArgsAndCols` was called on the original SQL — before the
rewrite — so PostgreSQL's `PREPARE` expanded `.*` in ordinal (creation)
order and returned `result_types` in that order. The parser then zipped
`source.columns` (alphabetical) with `resTypes` (ordinal) by position,
assigning the wrong Go type and nullable status to each column.

## Fix

Thread `columnOrder` into the parser and walker. In `getStarColumns`,
sort the columns alphabetically before writing the SQL expansion when
`columnOrder` is `"name"` — so the rewritten SQL reflects the sorted
order. Move `getArgsAndCols` to run on the formatted (already-rewritten)
SQL rather than the original input. PostgreSQL then prepares the sorted
expansion and returns `result_types` in the same order as
`source.columns`, keeping the zip correct.

Also aligns `column_order` pipe widths in the psql, mysql, and sqlite
driver docs tables.

---
*🤖 Created by agent*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Fixed column_order on `*` selectors in sql queries in postgres
+
 ### Added
 
 - Added `bob.NextUniqueInt()` for generating unique SQL alias suffixes ([#719](https://github.com/stephenafamo/bob/issues/719)). (thanks @atzedus)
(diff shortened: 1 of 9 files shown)


cbarber:
I can split these out into separate PRs if you prefer to squash merge over rebase merge.

PR #721, opened 2026-07-02
fix: replace randInt with bob.NextUniqueInt for join aliases (closes #719)

### LINKED ISSUE #719: randInt overflow bug
https://github.com/stephenafamo/bob/blob/bf3b99b5ce8fd16407da91f8c42d6cc178eb75e0/gen/templates/joins/bob_joins.bob.go.tpl#L66-L74

An unlikely but possible bug in the `randInt()` function.

`Sum64()` could potentially return `9223372036854775808` — in which case the output will be a negative number. There appears to be protection against this, but it doesn't work. Reproduction example:

```go
out64 := uint64(9223372036854775808)
out := int64(out64)
if out < 0 {
    fmt.Println(-out % 10000) // -5808
}
fmt.Println(out % 10000) // -5808
```

I would like to suggest removing `randInt` and switching to:
https://github.com/stephenafamo/bob/blob/bf3b99b5ce8fd16407da91f8c42d6cc178eb75e0/internal/utils.go#L58

This would also eliminate potential alias collisions caused by the overlap between `NextUniqueInt` and `randInt`.

However, `NextUniqueInt` is currently in the `internal` package, so it would need to be moved somewhere it can be imported. The question is where? `bob.NextUniqueInt()`?

atzedus:
Move NextUniqueInt from internal to bob as uint64 so generated join mods and preloaders share one counter without int64 overflow or collisions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Added `bob.NextUniqueInt()` for generating unique SQL alias suffixes ([#719](https://github.com/stephenafamo/bob/issues/719)). (thanks @atzedus)
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


atzedus:
Hi @stephenafamo,

Would you be able to cut a new release that includes this change? Our corporate security scanner is flagging the randInt overflow issue (#719), and we’d like to pick up the fix via a tagged version so we can clear the finding and move forward.

Appreciate your help!

stephenafamo:
Done. Released `v0.49.0`

PR #722, opened 2026-07-05
generate Has{Rel} EXISTS semi-join filters on the Where struct

sandonemaki:
## Summary

This PR adds generated `Has{Rel}(filters ...bob.Mod[*dialect.SelectQuery])` methods under the `R` field of the generated `{table}Where` struct. Each method filters parent rows by the **existence** of related rows, using a correlated `EXISTS` subquery (semi-join) built from the schema's relationship definitions.

```go
// "pilots that have a jet based at this airport"
models.SelectWhere.Pilots.R.HasJets(
    models.SelectJoins.Jets.InnerJoin.Airport,
    models.SelectWhere.Airports.ID.EQ(airportID),
)
```

```sql
WHERE EXISTS (
    SELECT 1 FROM "jets"
    INNER JOIN "airports" ON ...
    WHERE "jets"."pilot_id" = "pilots"."id"
      AND "airports"."id" = $1
)
```

## Motivation

Today, the only *generated* way to filter a parent table by a relationship is `SelectJoins` (INNER JOIN). This works, but multiplies parent rows by the number of matching children, which forces `DISTINCT`, silently inflates `Count()`, and breaks `LIMIT`-based pagination.

The `EXISTS` semi-join is the standard answer to this, and since v0.43 Bob has the `Exists` starter (thanks @manhrev). However, using it still requires hand-writing the correlation predicate with string-based expressions:

```go
sm.Where(psql.Exists(psql.Select(
    sm.Columns(psql.Raw("1")),
    sm.From("jets"),
    sm.Where(psql.Quote("jets", "pilot_id").EQ(psql.Quote("pilots", "id"))), // hand-written, not type-safe
    /* extra filters */
)))
```

The schema already knows this correlation — it's the same information `SelectJoins` uses to build `ON` clauses. This PR applies the same automation to `EXISTS`:

| hand-written | generated | what gets automated |
|---|---|---|
| `sm.InnerJoin(...).On(...)` | `SelectJoins.X.InnerJoin.Y` | the ON clause |
| `psql.Quote("t","col").EQ(v)` | `SelectWhere.X.Col.EQ(v)` | column name + type |
| `psql.Exists(hand-built subquery)` | `Has{Rel}(filters...)` **(this PR)** | the correlated subquery |

`Has{Rel}` is built **on top of** the `Exists` starter, not as an alternative to it: relationship-derived semi-joins are generated, free-form `EXISTS` remains the starter's territory — the same division of labor Bob already has between `SelectJoins` and `sm.InnerJoin`.

This is also a well-established pattern in the ecosystem: Ent generates `HasXWith(...)`, Laravel's `whereHas` produces exactly this correlated-EXISTS shape, Prisma has `some`.

## What is generated

For every relationship of a table, a method on the `R` sub-struct of the `{table}Where` struct:

```go
func (w pilotWhereR[Q]) HasJets(filters ...bob.Mod[*dialect.SelectQuery]) mods.Where[Q]
```

- **Return type is a plain `mods.Where[Q]`**, so it composes with everything else: it can be passed to `Query()`, combined with `WhereOr`/`WhereAnd`, and works for SELECT/UPDATE/DELETE via the `Q` type parameter.
- **`filters` are applied to the subquery**, so callers can join further and filter the related table using existing generated mods (see example above). Called with no arguments it is a pure existence check.
- **To-one, to-many, and many-to-many** are all covered. Single-side relationships share one code path (FK direction is absorbed by `FromColumns`/`ToColumns`); multi-side relationships build the subquery `FROM` the first side, `INNER JOIN` the remaining sides *inside* the subquery, and correlate only the first side back to the parent — so even m2m never multiplies parent rows.
- **`from_where`/`to_where` static conditions** on relationship sides are honored, consistent with the join/loader generators.
- **Self-referential relationships** alias the subquery's copy of the parent table, so the correlation keeps targeting the outer row instead of the subquery's own row.

## Design decisions

- **Naming (`Has{Rel}`)**: names the intent (has related rows), not the mechanism (EXISTS), and matches Ent's `HasX` vocabulary familiar to Go developers. Happy to rename if you prefer different wording.
- **Single method instead of `Has{Rel}` + `Has{Rel}With`**: the variadic `filters` covers both the bare existence check and the filtered case; splitting them adds API surface without adding capability.
- **Placement under `Where.R`**: the helper is semantically a WHERE-clause builder and needs the parent's (possibly aliased) column set, which the `Where` struct already carries — `AliasedAs` keeps working for free. Following review feedback, the methods live on an `R` sub-struct (mirroring the model struct's `R`) so they cannot collide with column-based filter fields, and a single alias resolves any collision for both.
- **Out of scope (possible follow-ups)**: a negated `NOT EXISTS` variant, and count-based predicates (`HAVING COUNT(*) >= n`).

## Testing

Adds a generated test file that, for every table with relationships, builds a query using each `R.Has{Rel}()` helper and asserts that it builds without error and emits an `EXISTS` clause. This runs across all dialects (psql, mysql, sqlite) since the templates are shared. Existing lint (`golangci-lint`) and `go test ./...` pass locally for all three dialects.

We have been running this generator in production against a PostgreSQL schema that includes to-one, has-many, many-to-many, and self-referential relationships.

## Checklist


stephenafamo:
Kindly update the changelog

sandonemaki:
@stephenafamo 
Added a changelog entry for this PR under Unreleased → Added.

stephenafamo:
My only issue with this is that it can potentially conflict with a column name.

Perhaps the best way to avoid this is to make relationship-based filters live under `R` similar to the main struct.  
This way if a table has a column named `r`, a single alias will prevent the collision both for the model struct field and the filter fields.

By this I mean something like:

```go
// "pilots that have a jet based at this airport"
models.SelectWhere.Pilots.R.HasJets(
    models.SelectJoins.Jets.InnerJoin.Airport,
    models.SelectWhere.Airports.ID.EQ(airportID),
)
```

sandonemaki:
@stephenafamo 
Done — moved the relationship filters under R as you suggested, mirroring the model struct: models.SelectWhere.Pilots.R.HasJets(...). The Where struct now carries an R sub-struct holding the Has{Rel} methods, so they can no longer collide with column-based filter fields. Also merged main into the branch.

PR #736, opened 2026-07-12
perf(orm): share preloaded to-one instances across parent rows

sandonemaki:
## Summary

When a to-one relationship is loaded with Preload (LEFT JOIN), the related
struct used to be built as a separate identical instance for every parent row
that references the same related row. Preload now deduplicates these per
execution: related rows with the same join key are shared as a single
instance — the same semantics as ThenLoad — without changing the emitted SQL
at all.

As discussed in the comments, this is Preload's standard behavior, not an
opt-in option: there is no new API surface and no alternative code path to
maintain.

## Motivation

In a production workload where 61,773 parent rows share 49 related rows
(~1,261x duplication), Preload builds and retains 60k+ identical instances.
This wastes both construction cost and retained memory, and nested ThenLoads
also expand the duplicate keys into their IN clauses as-is.

## Implementation (1 file, `orm/load.go`, ~+360 lines; plus ~550 lines of tests, docs, and a CHANGELOG entry)

- On the reflection `StructMapper` path, the join key is checked *before* the
  struct is built; on a cache hit, construction is skipped and the existing
  instance is shared
- With a typed/custom mapper the key is only readable *after* construction,
  so construction still runs per row; the duplicate is discarded in favor of
  the cached instance, and Collects are deduplicated
- A last-key fast path (one comparison, for sorted results) in front of a
  per-execution key-to-instance map
- As a safety valve for workloads with no duplication, dedup turns itself off
  when the first 1024 distinct keys see zero hits
- Automatically skipped (not an error) when the join key columns are not
  selected due to `PreloadOnly` / `PreloadExcept`; those preloads keep
  per-row instances

## Benchmarks (N=10,000 rows, median of 5 runs, vs the previous per-row behavior)

| Shape | Time | B/op | allocs/op |
|---|---|---|---|
| High duplication (50 distinct / 10,000 rows), keys sorted | −28% | −18% | −17% |
| High duplication, keys scattered | −26% | −18% | −17% |
| Zero duplication (worst case; auto-off engages) | +6% | +3% | +1% |

The memory savings (B/op, allocs/op) reproduce deterministically; treat the
time numbers as approximate since they are sensitive to environment noise.
With a custom (non-reflection) mapper, expect +5–10% time on duplicated
shapes in exchange for instance sharing and deduplicated Collects.

## Behavior change & tests

- The emitted SQL is unchanged. The observable change is that identical
  related rows become the same pointer (matching the existing semantics of
  ThenLoad) and construction / nested collection run once per distinct row.
  Preloaded structs should be treated as read-only; this is documented in the
  Preload docs and noted in the CHANGELOG.
- `orm/load_dedup_test.go` (~480 lines): instance sharing (both the
  reflection and typed mapper paths), non-consecutive duplicates, automatic
  shut-off, fallback when the key columns are not selected, composite keys,
  NULL keys, and nested Preload consistency against a baseline. All dialect
  test suites pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `NewViewx` and `NewTablex` in the `psql`, `mysql` and `sqlite` dialects now take a `scan.Mapper[T]` argument used for all queries built from the view/table. Pass `nil` to keep the previous reflection-based `scan.StructMapper` behaviour. `NewView`/`NewTable` are unchanged. (thanks @sandonemaki)
+- `Preload` now shares a single related-struct instance between all parent rows with the same join key, instead of building an identical instance per parent row — the same semantics as `ThenLoad`. The emitted SQL is unchanged; construction, nested loader collection and mapping now run once per distinct related row, which significantly reduces time, allocations and retained memory on high-duplication workloads. Preloaded structs should be treated as read-only: code that mutates a preloaded struct and relies on each parent having its own copy must copy the struct first. (thanks @sandonemaki)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `NewViewx` and `NewTablex` in the `psql`, `mysql` and `sqlite` dialects now take a `scan.Mapper[T]` argument used for all queries built from the view/table. Pass `nil` to keep the previous reflection-based `scan.StructMapper` behaviour. `NewView`/`NewTable` are unchanged. (thanks @sandonemaki)
+- JOIN a DISTINCT unnest(...) for composite-key relationship loaders & counts on PostgreSQL. (thanks @sandonemaki)
 - `Preload` now shares a single related-struct instance between all parent rows with the same join key, instead of building an identical instance per parent row — the same semantics as `ThenLoad`. The emitted SQL is unchanged; construction, nested loader collection and mapping now run once per distinct related row, which significantly reduces time, allocations and retained memory on high-duplication workloads. Preloaded structs should be treated as read-only: code that mutates a preloaded struct and relies on each parent having its own copy must copy the struct first. (thanks @sandonemaki)
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Fixed column_order on `*` selectors in sql queries in postgres
+
 ### Added
 
 - Added `bob.NextUniqueInt()` for generating unique SQL alias suffixes ([#719](https://github.com/stephenafamo/bob/issues/719)). (thanks @atzedus)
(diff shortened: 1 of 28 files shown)


sandonemaki:
@stephenafamo 
When you have a moment, we'd appreciate a review on this one. The per-row mapping cost it removes is something we're actively feeling in production, so if the approach looks good to you we'd be grateful for a merge — happy to address any feedback quickly, of course. CI is green and the branch is up to date with main.

stephenafamo:
In general, I dislike alternatives, I think it makes maintenance harder.

Is there a case against this being the only way preload works instead of being opt-in?

sandonemaki:
@stephenafamo 
No case against it. We'll update the PR: dedup becomes Preload's standard behavior, and the option is removed.

sandonemaki:
@stephenafamo 
Done — dedup is now Preload's only behavior. The PreloadDedup option and the dialect wrappers are removed, so there's no new API surface and a single code path. The read-only semantics are documented in the Preload docs and noted in the CHANGELOG, and the PR title/description are updated to match. All tests pass.

Please take a look and merge if you're happy with it.

stephenafamo:
Some conflicts

sandonemaki:
@stephenafamo 
Resolved the conflict with main (both branches had added a CHANGELOG entry in the same place) and merged the latest main into the branch. Build and tests pass.

stephenafamo:
I'm trying to review this but there are quite a number of new functions in `orm/load.go` and I would need time to properly understand the implementation.

I'm quite a bit busy at the moment, so it may take a while to find time to slowly review this

sandonemaki:
@stephenafamo 
Totally understand. To make the review cheaper, here is a map of the new code in `orm/load.go`. Nothing below changes any code; it's only a reading guide.

## The 10 new functions are 5 groups

| # | Group | Functions | What it does |
|---|---|---|---|
| 1 | Path selection | the mapper-mod returned by `buildPreloader` | picks one of three paths per execution: dedup-by-link (reflection mapper), dedup-by-value (typed mapper), or `legacy` |
| 2 | Dedup loops | `preloadDedupByLink`, `preloadDedupByValue` | the per-row logic; everything else serves these two |
| 3 | Cache | `preloadDedupCache` (`lookup`/`store`) | last-key fast path in front of a `map[string]T`, plus the auto-off valve |
| 4 | Key resolution | `preloadKeyColumnsSelected`, `preloadKeyFromLink`, `preloadKeyFromValue` | build a join-key extractor at mapper-setup time; return `nil` when they can't |
| 5 | Key encoding | `appendPreloadKeyValue`, `appendPreloadKeyPointer`, `appendPreloadKeyLen` | turn key column values into map-key bytes |

Suggested reading order: 1 → 2 → 3 → 4 → 5. Groups 4 and 5 are the largest but are pure functions with no state.

## Invariants that hold everywhere

- **Every failure path falls back to `legacy`**, which is the pre-PR per-row code extracted verbatim into a closure. A failed or skipped dedup can only mean "no dedup this row/execution", never different results. This covers: key columns dropped by `PreloadOnly`/`PreloadExcept`, a non-`*struct` mapper type, an unexpected link shape, and the auto-off valve.
- **NULL join keys (unmatched LEFT JOIN rows) are never cached** and always take the legacy behavior — checked via `wrapper.IsNull` on the link path and `rv.IsNil()` on the value path.
- **All state is per-execution.** The cache is created inside the mapper-mod generator, which `scan` runs once per query execution on a single goroutine — nothing is shared across queries, no locks.
- **On a cache hit, `Collect` is skipped**, so nested ThenLoads receive each distinct instance exactly once (this is what shrinks their `IN` clauses).

## Things you might flag, pre-answered

- **Why two dedup loops instead of one?** With the reflection `StructMapper`, the join key is readable from the link *before* the struct is built, so a duplicate row skips construction entirely (`preloadDedupByLink`). A typed/custom mapper's link is opaque — the key is only readable *after* construction — so we build, then discard the duplicate (`preloadDedupByValue`). Merging them would put the has-a-value-already conditional on every row of the common path.
- **`preloadKeyFromLink` peeks at scan's link representation** (`[]reflect.Value`, one entry per mapped column). It recomputes the expected length and the extractor checks `len(vals) != linkLen` on every row; any mismatch — including a future change inside `scan` — degrades to `legacy` per row rather than mis-keying.
- **The `%v` default in `appendPreloadKeyPointer` looks like it could print a pointer address.** It's unreachable from the StructMapper path: `wrapper.V` always holds a non-nil pointer from `reflect.New`, which the `rv.Kind() == reflect.Pointer && !rv.IsNil()` branch handles by recursing on the pointee.
- **Composite-key framing** (`data + 0x00 + decimal-length + 0x1f`) is injective: the encoding is parsed deterministically from the end (final `0x1f`, digits back to `0x00` give the part length), so part data containing the delimiter bytes cannot create collisions.
- **Why the last-key fast path?** Result sets are usually ordered, so consecutive rows share a key; one `bytes.Equal` replaces a map lookup on the hot path. **Why the auto-off at 1024?** It bounds the zero-duplication worst case to the +6% time / +3% B/op in the benchmark table.

If it helps, I can move the new functions into an `orm/load_dedup.go` so `load.go` stays close to what it was — a pure file move, no behavior or performance change. Otherwise I'll leave the diff untouched so nothing you've already read is invalidated. Happy to answer any question about a specific function here in the thread.

sandonemaki:
@stephenafamo 
Fully understand you're busy — no pressure on the depth of the review.

I wanted to share a bit of context on our side: we're planning to ship a release in early September that depends on this change, so a merge by the end of August would help us a lot. Would that timing be realistic for you? If so, that would be great — and if not, could you let me know roughly when a merge might be feasible?

Of course, I'm glad to answer any question about a specific function right here in the thread.

stephenafamo:
I'll find time to work review this in the next couple weeks, so a merge should be likely before August.

stephenafamo:
@sandonemaki I've released @v0.50.0  
I hope the state of things are sufficient for your release

PR #737, opened 2026-07-12
fix: close rows in QueryStmt.One to stop leaking a connection per call

sandonemaki:
## Summary

`QueryStmt.One` (the prepared-statement query path) leaks the `sql.Rows` —
and with `database/sql`, the pool connection holding them — on **every call**.

`One` passes the rows from `stmt.QueryContext` to `scan.OneFromRows`, which by
contract does not close caller-owned rows (`scan.One` does, `scan.OneFromRows`
does not), and `One` never closes them itself:

```go
rows, err := s.stmt.QueryContext(ctx, args...)
if err != nil {
    return t, err
}

t, err = scan.OneFromRows(ctx, s.mapper, rows)   // rows never closed
```

Because `One` reads a single row and never exhausts the result set,
`database/sql` cannot auto-release the connection. Each call permanently
holds one pool connection; a service using a prepared `One` in a request
handler exhausts the pool and PostgreSQL starts rejecting connections:

```
pq: sorry, too many clients already (53300)
```

This reproduces on every released version (verified on v0.42.0 and main)
against a real PostgreSQL 17 with a plain `bob.PrepareQueryx(...).One(...)`
loop.

## Changes

- `QueryStmt.One`: close the rows as soon as `scan.OneFromRows` returns
  (before the extra loaders run, matching the ordering of the non-prepared
  exec path where `scan.One` closes rows before `bob`'s loaders execute).
  A close error is returned if scanning itself succeeded.
- `QueryStmt.All`: the happy path exhausts the rows so `database/sql`
  auto-releases the connection, but a scan error mid-iteration leaked them —
  now closed explicitly on every path.
- `QueryStmt.Cursor`: rows ownership moves to the returned cursor, but when
  `scan.CursorFromRows` itself fails the rows leaked — now closed on that
  error path.
- `stmt_test.go` (new): regression tests using a fake `Preparer`/`Rows` that
  record whether `Close` was called — happy path, `sql.ErrNoRows` path,
  `All` happy path and `All` scan-error path. All four fail without this fix.

## Testing

- The four new regression tests fail on current main and pass with the fix.
- End-to-end: on a `MaxOpenConns(1)` pool, 10 consecutive `One` calls block
  forever without the fix (the leaked connection is never returned) and pass
  with it.
- `go test .` / `go vet .` pass.

### DIFF
--- a/stmt.go
+++ b/stmt.go
@@ -155,6 +155,9 @@ func (s QueryStmt[Arg, T, Ts]) One(ctx context.Context, arg Arg) (T, error) {
 	}
 
 	t, err = scan.OneFromRows(ctx, s.mapper, rows)
+	if closeErr := rows.Close(); err == nil {
+		err = closeErr
+	}
 	if err != nil {
 		return t, err
 	}
@@ -182,6 +185,9 @@ func (s QueryStmt[Arg, T, Ts]) All(ctx context.Context, arg Arg) (Ts, error) {
 	}
 
 	rawSlice, err := scan.AllFromRows(ctx, s.mapper, rows)
+	if closeErr := rows.Close(); err == nil {
+		err = closeErr
+	}
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


stephenafamo:
I feel like this is actually a bug in `scan` and should be fixed there

dmakushin:
> I feel like this is actually a bug in `scan` and should be fixed there

Please, review the fix - https://github.com/stephenafamo/scan/pull/10

stephenafamo:
Merged https://github.com/stephenafamo/scan/pull/10 and I've updated the scan package here.

I believe this PR can then be closed safely

PR #740, opened 2026-07-19
perf(gen): de-duplicate non-unique key arrays for ​= ANY($1)​ loaders & counts on PostgreSQL

sandonemaki:
> Follow-up to #714 (`col = ANY($1)`) and #718 (composite keys via `DISTINCT unnest`).
> This PR trims the **bound argument** of the single-column form; the SQL text is unchanged.

## Summary

Generated single-column relationship Slice loaders
(`gen/templates/models/table/009_rel_query.go.tpl`) and batch counts
(`gen/templates/counts/table/115_counts.go.tpl`) bind one key per parent to
`= ANY($1)`:

```go
pkGroupID := make(pgtypes.Array[int64], 0, len(os))
for _, o := range os {
    if o == nil {
        continue
    }
    pkGroupID = append(pkGroupID, o.GroupID) // one entry per parent, duplicates included
}
```

For a **to-one** relationship the keys are foreign key values, so parents that
share a related row all contribute the same key. One of our production shapes:
10,000 `tasks` rows pointing at 50 `groups` send a 10,000-element `bigint[]`
(~120KB on the pgx binary wire format) where 50 distinct keys (~0.6KB) produce
the exact same result — the array is only a semi-join filter, so duplicates
change nothing about the rows that come back.

This PR de-duplicates the array with a seen-map before binding it:

```go
pkGroupID := make(pgtypes.Array[int64], 0, len(os))
// the array is only a filter (semi-join), so duplicate keys can be
// dropped before they are sent over the wire
seenGroupID := make(map[int64]struct{}, len(os))
for _, o := range os {
    if o == nil {
        continue
    }
    if _, ok := seenGroupID[o.GroupID]; ok {
        continue
    }
    seenGroupID[o.GroupID] = struct{}{}
    pkGroupID = append(pkGroupID, o.GroupID)
}
```

To be clear about the relationship to #718's review discussion: as concluded
there, the semi-join does **not** need de-duplication for correctness. This
change is purely about the bytes bound to the query and the size of the hash
the planner builds on the DB side.

## When the seen-map is generated (and when it is not)

The map is only worth its cost when the key column can actually contain
duplicates, so codegen gates it on the schema:

| key column | example | generated |
| --- | --- | --- |
| non-unique FK (to-one, high fan-in possible) | `videos.user_id` | **seen-map dedup** |
| parent's own PK (to-many loaders, `LoadCount<Rel>`) | `videos.id` | plain loop (unchanged) |
| uniquely-constrained FK (one-to-one) | `videos.sponsor_id unique` | plain loop (unchanged) |
| type not comparable with `==` | custom `compare_expr` types | plain loop (unchanged) |
| composite keys | — | unchanged (#718's `DISTINCT unnest` already de-duplicates in SQL) |
| non-PostgreSQL dialects | — | unchanged (row-value `IN` list) |

Uniqueness is checked with the existing `drivers.Table.HasExactUnique` (primary
key or unique constraint), and `==`-comparability with `Types.CanCompareWithEquals`
— the same gating as the `copyMatchingRows` map-stitch (#729).

## Why the uniqueness gate

Benchmarking the two generated fill-loop shapes at 10,000 parents:

```
no dedup (before)               53µs    82KB/op
seen-map, fan-in 50 (after)    455µs   377KB/op   → binds 50 keys instead of 10,000
seen-map on unique keys      1,256µs               → zero wire savings
```

On keys that are unique by construction the map is ~1.2ms of pure overhead per
10,000 parents, so those call sites keep the previous code — the gate is decided
at codegen time and costs nothing at runtime.

## Also included: two small cleanups in the through-loader template

`gen/templates/loaders/table/110_loaders.go.tpl` (`Load<Rel>` for through
relationships) applied every query mod **twice**: once against a throwaway
`SelectQuery` just to check whether the user set any columns, then again for
real:

```go
sq := dialect.SelectQuery{}
for _, mod := range mods {
    mod.Apply(&sq)          // full apply, result discarded
}
if len(sq.SelectList.Columns) == 0 {
    mods = append(mods, sm.Columns(Tags.Columns))
}
q := os.Tags(append(mods, ...)...)  // full apply again
```

The throwaway pass is replaced with a deferred `bob.ModFunc` placed after the
user mods and before the loader's `related_*` columns, so it makes the same
decision at the same point during the single real application:

```go
q := os.Tags(append(
    mods,
    bob.ModFunc[*dialect.SelectQuery](func(q *dialect.SelectQuery) {
        if len(q.SelectList.Columns) == 0 {
            sm.Columns(Tags.Columns).Apply(q)
        }
    }),
    sm.Columns(VideoTags.Columns.VideoID.As("related_videos.ID")),
)...)
```

This is the same deferred-default-columns pattern `View.Query` itself already
uses (`dialect/psql/view.go`), and it sees exactly what the old check saw: the
`SelectQuery` starts with no columns and, at that position in the mod list, only
the user's mods have been applied.

The same template also grows the join-key slice from zero capacity; it is now
pre-allocated to `len(os)` as a lower-bound hint.

## Behavior (no functional change)

- **Results are identical.** The array only filters (`WHERE key = ANY($1)`);
  duplicates in it cannot multiply rows, and counts group over child rows, so
  `LoadCount` values are unchanged too.
- **NULL semantics preserved** — `= ANY` never matches NULL; at most the array
  now carries one NULL instead of many.
- **SQL text is unchanged** — only the bound array shrinks, so prepared
  statements and plans are unaffected.
- **Column-defaulting decision unchanged** — verified against `View.Query`'s
  mod-application order (see above).

## Verification

- Generated against the test schema plus a composite-FK pair
  (`comp_children(parent_region, parent_code) → comp_parents`): dedup appears
  exactly on non-unique FK keys (`videos.user_id`, self-referencing FKs,
  join-table FKs), and **not** on PK keys, on `videos.sponsor_id` (unique), or
  on composite keys; everything compiles (`go build ./models/...`).
- `go test ./gen/bobgen-psql/...` — full suite (golden files + generated-code
  tests against a real PostgreSQL) passes.
- `go test ./gen/bobgen-sqlite/...` (modernc/ncruces) passes — the through-loader
  change is dialect-independent, non-psql loader output is otherwise unchanged.
- Fill-loop micro-benchmarks above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `NewViewx` and `NewTablex` in the `psql`, `mysql` and `sqlite` dialects now take a `scan.Mapper[T]` argument used for all queries built from the view/table. Pass `nil` to keep the previous reflection-based `scan.StructMapper` behaviour. `NewView`/`NewTable` are unchanged. (thanks @sandonemaki)
 - JOIN a DISTINCT unnest(...) for composite-key relationship loaders & counts on PostgreSQL. (thanks @sandonemaki)
+- PostgreSQL single-column slice relationship loaders (`<Parent>Slice.<Rel>`) and batch counts (`<Parent>Slice.LoadCount<Rel>`) now de-duplicate the key array bound to `= ANY($1)` when the key column can contain duplicates (a non-unique foreign key). The array is only a semi-join filter, so query results are unchanged — a slice of 10,000 parents sharing 50 related rows now binds 50 keys instead of 10,000. Keys that are unique by construction (the parent's own primary key or a uniquely-constrained column) and key types not comparable with `==` keep the previous plain loop, decided at codegen time ([#740](https://github.com/stephenafamo/bob/pull/740)). (thanks @sandonemaki)
+- Generated through-relationship loaders (`Load<Rel>`) no longer apply every query mod twice (once against a throwaway query to detect whether the user set columns, then again for real). The default columns are now added by a deferred `bob.ModFunc` during the single real application — the same pattern `View.Query` uses — and the join-key slice is pre-allocated to the parent slice length. Generated SQL is unchanged ([#740](https://github.com/stephenafamo/bob/pull/740)). (thanks @sandonemaki)
 
 ### Fixed
 
--- a/gen/templates/counts/table/115_counts.go.tpl
+++ b/gen/templates/counts/table/115_counts.go.tpl
@@ -193,6 +193,39 @@ func (os {{$tAlias.UpSingular}}Slice) LoadCount{{$relAlias}}(ctx context.Context
 	{{- $fromCol := index $firstFrom.Columns $local}}
 	pk{{$fromCol}} := make(pgtypes.Array[{{$colTyp}}], 0, len(os))
 	{{- end}}
+	{{- if eq (len $firstSide.FromColumns) 1}}
+	{{- $local := index $firstSide.FromColumns 0 -}}
+	{{- $column := $.Table.GetColumn $local -}}
+	{{- $colTyp := $.Types.GetNullable $.CurrentPackage $.Importer $column.Type $column.Nullable -}}
+	{{- $fromCol := index $firstFrom.Columns $local}}
+	{{- /* keys that are unique by construction (the parent's own PK or a
+	       unique column) never contain duplicates, so the seen-map would be
+	       pure overhead; dedup only de-duplicatable, ==-comparable keys */ -}}
+	{{- $canDedup := and (not ($.Table.HasExactUnique $local)) ($.Types.CanCompareWithEquals $.CurrentPackage $column.Type)}}
+	{{- if $canDedup}}
+	// the array is only a filter (semi-join), so duplicate keys can be
+	// dropped before they are sent over the wire
+	seen{{$fromCol}} := make(map[{{$colTyp}}]struct{}, len(os))
+	for _, o := range os {
+		if o == nil {
+			continue
+		}
+		if _, ok := seen{{$fromCol}}[o.{{$fromCol}}]; ok {
+			continue
+		}
+		seen{{$fromCol}}[o.{{$fromCol}}] = struct{}{}
+		pk{{$fromCol}} = append(pk{{$fromCol}}, o.{{$fromCol}})
+	}
+	{{- else}}
+	for _, o := range os {
+		if o == nil {
+			continue
+		}
+		pk{{$fromCol}} = append(pk{{$fromCol}}, o.{{$fromCol}})
+	}
+	{{- end}}
+	PKArgExpr := {{$.Dialect}}.Any({{$.Dialect}}.Cast({{$.Dialect}}.Arg(pk{{$fromCol}}), "{{$column.DBType}}[]"))
+	{{- else}}
 	for _, o := range os {
 		if o == nil {
 			continue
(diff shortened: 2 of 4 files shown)


stephenafamo:
Wouldn't this be worse for smaller lists?

sandonemaki:
@stephenafamo 
> Wouldn't this be worse for smaller lists?

I benchmarked this against the workload the 10k-rows/50-groups example came from (a list endpoint: an `items` slice → to-one `groups` loader), on a real PostgreSQL. Both variants run the exact same builder and query — `SELECT ... FROM groups WHERE id = ANY($1::bigint[])`; the only difference is the dedup loop before binding. I compared the v0.48 generated `ItemSlice.Group()` verbatim vs this PR's generated shape.

**End-to-end loader query (array build + execute + scan), median of 5 runs:**

| N (parents) | distinct keys | before | after | Δ |
|---:|---:|---:|---:|---:|
| 10 | 10 (all unique) | 0.61 ms | 0.61 ms | ±0% |
| 100 | 100 (all unique) | 0.87 ms | 0.88 ms | +1% (noise) |
| 1,000 | 50 | 2.59 ms | 0.73 ms | **−72%** |
| 1,000 | 1,000 (all unique) | 5.33 ms | 5.37 ms | +0.8% |
| 10,000 | 50 | 5.41 ms | 0.97 ms | **−82%** |
| 10,000 | 500 | 7.30 ms | 5.43 ms | −26% |
| 10,000 | 10,000 (all unique) | ~46 ms | ~46 ms | ±0% |
| 40,000 | 50 | 17.0 ms | 1.9 ms | **−89%** |
| 40,000 | 2,000 | 21.1 ms | 10.6 ms | −50% |
| 40,000 | 40,000 (all unique) | 172.8 ms | 172.9 ms | ±0% |

For small lists the difference is unmeasurable in either direction — at N≤100 all configurations (duplicated or all-unique) were within run-to-run noise of a single ~0.5 ms round-trip.

The worst case for the dedup is all-unique values in a non-unique FK column: the map buys nothing and both variants send a byte-identical query, so the true added cost is exactly the array-build delta. Measured in isolation (no DB):

| N | plain append | dedup | added cost | relative to the query itself |
|---:|---:|---:|---:|---:|
| 10 | 55 ns (1 alloc) | 430 ns (4 allocs) | +0.4 µs | ~0.07% of 0.6 ms |
| 100 | 0.4 µs | 3.1 µs | +2.7 µs | ~0.3% of 0.9 ms |
| 1,000 | 4 µs | 32 µs | +28 µs | ~0.5% of 5.3 ms |
| 10,000 | 64 µs | 332 µs | +0.27 ms | ~0.6% of 46 ms |
| 40,000 | 0.44 ms | 1.76 ms | +1.3 ms | ~0.8% of 173 ms |

So the overhead ceiling stays under ~1% of the query it precedes at every size from 10 to 40,000, while any realistic duplication turns into a 1.3×–8.8× speedup (and the bind array shrinks accordingly, e.g. ~560 kB → ~0.7 kB at N=40,000/D=50 with the bigint ids used here). Note this is against a loopback Postgres — over a real network link the duplicate-array penalty only gets bigger, the dedup cost doesn't.

The uniqueness gate already keeps the map off the only columns where it could never pay for itself, so I believe this covers the concern — nothing further pending from my side, ready to merge whenever you are.

<details>
<summary>Setup</summary>

- Go 1.26.4, PostgreSQL 17.6 (Docker, loopback), linux/amd64, Intel Core Ultra 5 235U, `-cpu 1 -count 5` (medians)
- Parents are in-memory model instances with the FK assigned round-robin over D seeded child rows; child table has ~40k rows
- "before" runs the v0.48 generated slice-relationship method unmodified; "after" runs the same query with the key array deduplicated exactly as generated by this PR

</details>

PR #743, opened 2026-07-21
Fix doubled parentheses in Exists/Any/All with query operands

### LINKED ISSUE #742: Generated many-to-many Where filter emits invalid EXISTS ((SELECT ...)) with doubled parentheses
### What's wrong

The many-to-many `Where` helper generated by #722 produces SQL with doubled parentheses around the subquery — `EXISTS ((SELECT ...))` — which fails to execute (observed on SQLite). Parent queries that filter through a many-to-many relationship return an error instead of the matching rows.

The generated code builds a `SelectQuery` and returns `dialect.Exists(q)`. `Exists` wraps its operand in `group{}` unconditionally:

```go
// expr/builder.go
func Exists[T bob.Expression, B builder[T]](exp bob.Expression) T {
    var b B
    return b.New(Join{Exprs: []bob.Expression{exists, group{exp}}})
}
```

`group` always writes its own parentheses, and the query operand also renders parentheses on its own, so the two stack. Notably, the general `X()` builder already guards against this by checking `ShouldOmitParens` before wrapping:

```go
case bob.ParensOmitter:
    if t.ShouldOmitParens() { break }
    exp = group{exp}
```

but `Exists` (and `Any`/`All`) bypass that path and wrap unconditionally.

### Where

- `expr/builder.go` `Exists`: https://github.com/stephenafamo/bob/blob/main/expr/builder.go#L47-L50
- Generated template returning `dialect.Exists(q)`: https://github.com/stephenafamo/bob/blob/main/gen/templates/where/table/100_where.bob.go.tpl#L181-L183
- `X()` omit-parens guard for comparison: https://github.com/stephenafamo/bob/blob/main/expr/builder.go#L26-L34

### How to reproduce

1. Generate the `Where` API for a schema with a many-to-many parent to endpoint relationship.
2. Build a parent select using the generated `Parents.R.HasEndpoints(...)` helper.
3. Render/execute against SQLite and inspect the SQL — the `EXISTS ((SELECT ...))` form fails.

### Suggested fix

Route the `Exists`/`Any`/`All` operand through the same `ShouldOmitParens` check that `X()` uses (or drop the extra `group{}` when the operand already parenthesizes itself), so the output is `EXISTS (SELECT ...)`.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/41931ac6-18e0-4347-b2ee-956798c94b55.

evanmarshall:
Fixes #742

## The bug

`Exists`, `Any` and `All` in `expr/builder.go` wrapped their operand in `group{}` unconditionally. But a `bob.Query` operand (such as the `bob.BaseQuery` returned by `psql.Select`/`sqlite.Select`/`mysql.Select`) already writes its own surrounding parentheses when rendered as an expression (`bob.BaseQuery.WriteSQL`). The two stacked into invalid SQL like `EXISTS ((SELECT ...))`, which fails to execute on SQLite and broke the generated relationship `Where` filters from #722 (`Parents.R.HasEndpoints(...)` etc., which return `dialect.Exists(q)`).

## The fix

`Exists`/`Any`/`All` now skip the extra `group{}` when the operand implements `bob.Query`, relying on the query to write its own parentheses — the same assumption `bob.BaseQuery.WriteSQL` already makes for nested queries.

I deliberately did **not** reuse the `ShouldOmitParens` guard from `X()` as the issue first suggested: `Arg`, `Raw`, `Clause` etc. report `ShouldOmitParens() == true` but do not write their own parentheses, and `EXISTS`/`ANY`/`ALL` require parentheses around their operand — so routing through that check would have turned `ANY ($1)` into invalid `ANY $1`. Non-query operands keep the group exactly as before.

Also updated the doc comments in the three dialects' `starters.go`, which documented the doubled form (`EXISTS ((SELECT 1))`).

## Tests

- `expr/builder_test.go`: exact-string assertions that a query operand renders as `EXISTS (SELECT ...)` / `ANY (...)` / `ALL (...)` with a single set of parentheses, and that non-query operands (`Raw`, `Arg`) still get wrapped. Exact matching matters here because `testutils.Clean` collapses doubled parentheses, which would mask this regression.
- `dialect/sqlite/select_test.go`: an end-to-end `WHERE EXISTS (subquery)` case validated by the SQLite grammar parser.

`go test ./...` passes except the pre-existing `TestLibSQL` container failure in `gen/bobgen-sqlite/driver`, which fails identically on a clean checkout of `main` (Docker/testcontainers environment issue, unrelated).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/expr/builder.go
+++ b/expr/builder.go
@@ -46,7 +46,7 @@ func Not[T bob.Expression, B builder[T]](exp bob.Expression) T {
 // Exists expression
 func Exists[T bob.Expression, B builder[T]](exp bob.Expression) T {
 	var b B
-	return b.New(Join{Exprs: []bob.Expression{exists, group{exp}}})
+	return b.New(Join{Exprs: []bob.Expression{exists, subquery(exp)}})
 }
 
 // prefix the expression with a - (minus)
@@ -58,13 +58,26 @@ func Minus[T bob.Expression, B builder[T]](exp bob.Expression) T {
 // ANY expression
 func Any[T bob.Expression, B builder[T]](exp bob.Expression) T {
 	var b B
-	return b.New(Join{Exprs: []bob.Expression{anyOp, group{exp}}})
+	return b.New(Join{Exprs: []bob.Expression{anyOp, subquery(exp)}})
 }
 
 // ALL expression
 func All[T bob.Expression, B builder[T]](exp bob.Expression) T {
 	var b B
-	return b.New(Join{Exprs: []bob.Expression{all, group{exp}}})
+	return b.New(Join{Exprs: []bob.Expression{all, subquery(exp)}})
+}
+
+// A [bob.Query] writes its own surrounding parentheses when rendered as an
+// expression (see [bob.BaseQuery.WriteSQL]), so wrapping it in a group would
+// produce doubled parentheses such as EXISTS ((SELECT ...)).
+// Any other expression is wrapped in a group since EXISTS/ANY/ALL require
+// parentheses around their operand.
+func subquery(exp bob.Expression) bob.Expression {
+	if _, ok := exp.(bob.Query); ok {
+		return exp
+	}
+
+	return group{exp}
 }
 
 // To be embedded in query mods
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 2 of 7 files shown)


expr/builder.go
```diff
@@ -58,13 +58,26 @@ func Minus[T bob.Expression, B builder[T]](exp bob.Expression) T {
...
+// expression (see [bob.BaseQuery.WriteSQL]), so wrapping it in a group would
+// produce doubled parentheses such as EXISTS ((SELECT ...)).
+// Any other expression is wrapped in a group since EXISTS/ANY/ALL require
+// parentheses around their operand.
+func subquery(exp bob.Expression) bob.Expression {
+	if _, ok := exp.(bob.Query); ok {
+		return exp
+	}
+
+	return group{exp}
```
stephenafamo, line 80: This function name seems wrong... perhaps it should be named something like`groupIfNotQuery` or something clearer

stephenafamo (changes requested):

PR #745, opened 2026-07-27
fix(psql): handle duplicate parameter refs ($N) in query parsing

### LINKED ISSUE #744: Issue with v0.49.0 and generate queries
Seems like there is some problem when generating queries:
```
-- GetClient
select * from client where equip_id = $1 or browser = $2;
```
works

```
-- GetClient
select * from client where equip_id = $1 or browser = $1;
```
gives 
`unable to fetch table data: parse query folders: parse folder "./query": parse file "sys.sql": parse queries: parse query 0: expected 6 args, got 6`

dmakushin:
When a query uses the same placeholder ($1) multiple times, the parser incorrectly renumbered each occurrence sequentially, causing a mismatch between expected args from PostgreSQL PREPARE and the tracked args.

Added paramIdxMap to map each unique original param number to a single sequential index, ensuring duplicate $N references are collapsed into one parameter for type resolution.

Also fixed the error message at args mismatch to print the actual expected/got values (argTypes vs w.args) instead of misleading column counts.

Fixes #744


stephenafamo:
Add a note in CHANGELOG.md include thanks @yourhandle

stephenafamo:
@polderudo can you confirm that this fixes your issue?

polderudo:
I can confirm, that this fixes the error

PR #752, opened 2026-08-12
fix: avoid duplicate parentheses around subqueries

### LINKED ISSUE #742: Generated many-to-many Where filter emits invalid EXISTS ((SELECT ...)) with doubled parentheses

fzlzjerry:
## Summary

- avoid adding another group around query operands passed to `Exists`, `Any`, and `All`
- preserve the required grouping for ordinary expression operands
- add exact-output regression coverage and tighten the generated relationship-filter assertion

## Tests

- `go test -timeout 20m -race -covermode atomic -coverprofile=/tmp/bob-742-coverage.out -coverpkg=github.com/stephenafamo/bob/... ./...`
- `golangci-lint run`

Closes #742


fzlzjerry:
I missed #743 when checking for existing work. It already addresses the same root cause and has maintainer feedback, so I'm closing this as a duplicate rather than splitting review across two patches. Sorry for the noise.

PR #758, opened 2026-08-16
fix!: copy only the column fields in the generated single-model Update and Reload

sandonemaki:
## Summary

The generated single-model `(*Foo).Update` and `(*Foo).Reload` now copy **only the column fields** out of the row they fetched, instead of overwriting the whole receiver. Every non-column field on the model — the relationship cache `.R` (including `R.Loaded`), the counts plugin's `.C`, and anything a plugin adds through the `model/fields/additional` template block — is now left untouched.

## Problem

Both methods previously assigned a freshly scanned model over the receiver wholesale:

```go
*o = *v   // Update
*o = *o2  // Reload
```

`v` (from `UPDATE ... RETURNING`) and `o2` (from a plain `SELECT`) never load relationships or counts, so all of their non-column fields are zero values. The net effect was that a successful `Update()` or `Reload()` silently emptied `.R`, reset every `R.Loaded.<Rel>` flag to `false`, and cleared every count pointer in `.C`. `Reload` did carry `.R` forward explicitly (`o2.R = o.R` before `*o = *o2`), but nothing carried `.C` or any other plugin field, on any dialect. `Update` (default block, used by PostgreSQL/SQLite) carried neither.

This confirms the maintainer's earlier point that the previous `.R`-only stash/restore fix (this PR's original approach) missed the counts plugin's `.C` field entirely — it is a real, distinct bug, not a hypothetical one.

## Why "copy only the columns"

A first version of this PR fixed `.R` by stashing and restoring it around the wholesale assignment (`oldR := o.R` / `*o = *v` / `o.R = oldR`). That approach only protects the one field it names. The set of non-column fields on a generated model is **open-ended**: the counts plugin adds `.C`, and any third-party plugin can add more fields through the `model/fields/additional` template hook. A fix that enumerates what to *preserve* can never be complete — each new field source needs its own stash/restore line.

The set of **columns**, by contrast, is fully known at code-generation time (`$table.Columns`). So instead of enumerating what to preserve, this PR enumerates what to *copy*: assign each column field individually (`o.<Col> = v.<Col>` / `o.<Col> = o2.<Col>`) and touch nothing else. This is structurally complete — it protects `.R`, `.C`, and any future plugin field without adding a single line to the template when a new plugin field shows up.

The MySQL dialect already works this way: its `one_update` override calls `s.Overwrite(o)`, which only ever touches columns. This PR brings the default block (PostgreSQL, SQLite) in line for `Update`, and — since `Reload` has no dialect override — fixes `Reload` on every dialect, MySQL included.

## Scope

This PR is limited to `gen/templates/models/table/005_one_methods.go.tpl`'s `Update` (default block only; MySQL's `one_update` override is untouched, since it already copies columns only) and `Reload` (all dialects, since it has no override).

The slice-side path (`007_slice_methods.go.tpl`'s `copyMatchingRows`, used by `UpdateAll`/`ReloadAll`) has the same underlying problem (`new.R = old.R` is carried, `.C` is not) but its fix involves a genuine design choice between two different-shaped approaches, and is intentionally **not** included here. It is tracked as a separate, follow-up issue so this PR can stay focused on the single-model path.

## Generated columns

The range is over `$table.Columns` — **all** columns, including generated ones — not `NonGeneratedColumns` and not the `Overwrite` helper's skip-generated/skip-unset logic. `UPDATE ... RETURNING` (and the plain `SELECT` for `Reload`) return values computed by the database — generated columns, trigger results, `DEFAULT`s — and copying only non-generated columns would silently drop those values from `o`. Copying the full column set preserves the previous "trust what came back from the round trip" behavior exactly.

## Tests

The existing regression test template, `gen/templates/factory/bobfactory_update_related.bob_test.go.tpl` (`Test<Foo>UpdateKeepsLoadedRelationships`), needed no behavioral changes — only its docstring was updated to describe the new, broader contract. Under the new approach, `.R` and `.C` are protected by the exact same mechanism (no assignment statement targets them at all), so the existing `.R` test is a structural gate: any regression that reintroduces a wholesale assignment, or that widens the column range to copy non-column fields, would make this test fail.

We intentionally did **not** add a test that directly asserts `.C` survives `Update`/`Reload`. The factory-output template that would host such a test has no way to statically know whether the counts plugin is enabled for a given generator invocation — `TemplateData` carries no plugin-enablement flags, and counts is not a separate `output`, just extra templates appended to the `models` output. Writing `obj.C` directly into the test would fail to compile for any user who has the counts plugin disabled. A `reflect`-based test (looking up the `C` field by name, which degrades gracefully to a no-op when the field doesn't exist) would work without this problem, and we're happy to add it if maintainers would like — we left it out by default since generated tests using `reflect` tend to be less welcome in review.

## Breaking change

Any code relying on `Update()` or `Reload()` clearing `.R` — most notably a "reload if not loaded" pattern keyed off `R.Loaded.<Rel>` — will stop reloading, because `R.Loaded.<Rel>` now stays `true` and keeps pointing at pre-update data. The same applies to `.C`: counts loaded before the call now survive it, and are no longer implicitly invalidated. `.R` and `.C` have always been caches with no freshness guarantee; callers that need post-update relationship data or counts should fetch it explicitly (`Reload` for columns, `Load<Rel>` / `Preload` / `LoadCount<Rel>` / `PreloadCount` for the caches) rather than relying on these methods to clear the caches as a side effect.

## Checklist

- [x] Template change scoped to `005_one_methods.go.tpl`'s `Update` (default block) and `Reload` (all dialects); MySQL's `one_update` override is untouched.
- [x] Existing regression test template updated (docstring only); no new test code needed since the fix mechanism subsumes the existing test.
- [x] Updated CHANGELOG.md under `[Unreleased]` with a `BREAKING` note covering both `Update` and `Reload`, and both `.R` and `.C`.
- [x] Slice-side `copyMatchingRows` intentionally left out of scope; tracked separately.


stephenafamo:
I think this may also affect some other plugins (e.g. the Count plugin)
Perhaps there should be a template hook in the update method that plugins can write into, although I think you can only fully overwrite a template block

Perhaps the true fix is to copy ONLY the column fields into the new struct

sandonemaki:
You are right. I will handle it.

sandonemaki:
Thanks for the review — all three points landed, and I've reworked this PR accordingly.

**On point 1 (Counts plugin impact):** you were right, and the impact was actually broader than just `Update`. The wholesale `*o = *v` / `*o = *o2` assignments in the default block clear every non-column field, not just `.R`. That includes the counts plugin's `.C` (on `Update`, all dialects except MySQL, and on `Reload`, all dialects including MySQL, since `Reload` has no dialect override), and it would include any field a third-party plugin adds through the `model/fields/additional` hook.

**On point 3 (true fix: copy only the columns), adopted:** this PR now does exactly that. `Update` and `Reload` copy each column field individually (`o.<Col> = v.<Col>` / `o.<Col> = o2.<Col>`) instead of overwriting the receiver wholesale, so `.R`, `.C`, and any future plugin field are protected automatically, with zero additional lines needed per plugin. The set of columns is fully known at generation time; the set of non-column fields is open-ended, so this is the only approach that's structurally complete rather than needing a new stash/restore line every time a new non-column field source shows up. MySQL's `one_update` override already worked this way (`s.Overwrite(o)`), so this brings the default block in line for `Update`, and fixes `Reload` on every dialect since it has no override to begin with.

**On point 2 (template hook mechanism), not adopted:** I looked into this — it's feasible (bob already does something similar for `model/fields/additional` and `model/fields/counts`, both empty blocks a plugin can `{{define}}` into) but it has two structural drawbacks compared to the columns-only approach:

1. It needs a hook point in three places (`Update`, `Reload`, and the slice-side `copyMatchingRows`), and each of those needs *two* hook points (one before the assignment to stash, one after to restore) — six hook points total for one concern.
2. `text/template`'s `{{define}}` doesn't error on redefinition — the loader (`gen/output.go:178-186`) sorts template files by name and parses them in order, so if two plugins ever `{{define}}` the same hook name, the second one silently wins with no error or warning. In practice that limits each hook point to one plugin.

Given that the columns-only approach solves the same problem without a new mechanism, I went with that instead.

**Scope note:** this PR is limited to the single-model `Update` and `Reload` in `005_one_methods.go.tpl`. The slice-side `copyMatchingRows` (used by `UpdateAll`/`ReloadAll`) has the same underlying issue but its fix involves a genuine design choice between two different-shaped approaches (add a preserve-list vs. flip which side is copied into which), so I've split it out as a separate follow-up issue rather than block this PR on it.

I also intentionally did not add a test that directly asserts `.C` survives `Update`/`Reload` — the factory-output template has no way to statically know whether the counts plugin is enabled for a given generator run, so writing `obj.C` into a generated test would break compilation for anyone with counts disabled. The existing `.R` regression test now doubles as a structural gate for this fix, since `.R` and `.C` are protected by the exact same mechanism (no assignment statement targets either). Happy to add a `reflect`-based `.C` check if you'd like one — I left it out by default since `reflect` in generated test code tends to be unwelcome, but I don't feel strongly about it either way.

sandonemaki:
@stephenafamo 
I have addressed the points raised.
Please review the PR.
#758 
#759

stephenafamo (approved):

PR #759, opened 2026-08-17
fix: copy only the column fields in the generated slice copyMatchingRows

sandonemaki:
## Summary

`<Table>Slice.copyMatchingRows` — the helper behind `UpdateAll`, `ReloadAll` and the `UpdateMod`/`DeleteMod`/`MergeMod` loaders — carried only `.R` across to the freshly scanned model and then swapped the slice element to point at it. Every other non-column field, most visibly the counts plugin's `.C`, was silently reset to its zero value. This changes the helper to copy only the column fields onto the model already in the slice, leaving every non-column field untouched and keeping the slice's original pointers.

This is the slice-side counterpart of [#758](https://github.com/stephenafamo/bob/pull/758), which fixed the same class of bug in the single-model `Update` and `Reload`.

## Problem

Generated code for a table that has at least one relationship (`{{if $.Relationships.Get $table.Key}}` gates the `new.R = old.R` line, so tables without relationships never had it):

```go
for i, old := range o {
    new, ok := fromByPK[old.ID]
    if !ok {
      continue
    }
    new.R = old.R
    o[i] = new
}
```

`new` comes from scanning an `UPDATE ... RETURNING` or a plain `SELECT`, neither of which loads relationships or counts, so all of its non-column fields are zero values. Carrying `.R` over by hand covers exactly one of them. The counts plugin's `.C` — and anything a third-party plugin adds through the `model/fields/additional` block — was dropped on every call.

The set of non-column fields is open-ended: plugins add to it, so a fix that enumerates what to *preserve* can never be complete. The set of columns, by contrast, is fully known at generation time, so enumerating what to *copy* is complete by construction. That is the same reasoning behind #758.

## Affected paths

| Path | Dialects | Was `.C` lost? |
| --- | --- | --- |
| `UpdateAll` (default `slice_update` block) | PostgreSQL, SQLite | Yes |
| `ReloadAll` | **All** (no dialect override exists) | Yes |
| `MergeMod` used with `.All()` | PostgreSQL | Yes |
| `DeleteAll` | All | No — it runs `.Exec()`, so the loader receives `nil` and never reaches `copyMatchingRows` |
| `UpdateAll` (MySQL `slice_update` override) | MySQL | No — it calls `vals.Overwrite(o[i])` and never swaps pointers |

MySQL's override already copies columns onto the models already in the slice. This PR brings the default path in line with it.

## Fix

`gen/templates/models/table/007_slice_methods.go.tpl` now reverses the direction of the copy in both branches of `copyMatchingRows` (the map fast path and the nested-loop fallback for composite or non-`==`-comparable keys):

```go
for _, old := range o {
    new, ok := fromByPK[old.ID]
    if !ok {
      continue
    }
    old.ID = new.ID
    old.Name = new.Name
    // ... one assignment per column
}
```

All columns are copied, including generated ones, so values computed by the database still land on the model exactly as before. The `{{if $.Relationships.Get $table.Key}}` gate around the old `new.R = old.R` is gone, since no non-column field is touched any more.

## Breaking change

The models in the slice are no longer replaced with the freshly scanned ones. Code that relied on the pointer identity changing — for example comparing a previously held `*Foo` against `slice[i]` to detect a refresh — will now observe the two pointers as equal.

Conversely, and this is the intended improvement, a caller holding a pointer to a member of the slice now *does* see the new column values after `UpdateAll`/`ReloadAll`, which matches how the single-model `Update`/`Reload` already behave.

Additionally, the freshly scanned models are no longer mutated: previously `copyMatchingRows` wrote the old model's `.R` onto them. Callers that run `Table.Update(vals.UpdateMod(), slice.UpdateMod()).All(...)` (or the `MergeMod` equivalent) themselves and read `.R` off the returned slice will now find it empty; the loaded relationships stay on the models in the original slice instead.

## Regression test

Added `gen/templates/factory/bobfactory_reload_all_related.bob_test.go.tpl`, following the same shape as the test template added in #758. For each table with a primary key and at least one relationship, it generates a test that puts a known instance into `.R.<Rel>`, sets `R.Loaded.<Rel>`, calls `ReloadAll` on a one-element slice, and asserts that (a) the slice still holds the original pointer, (b) `.R` and `R.Loaded` survived, and (c) — for tables with at least one non-primary-key column — that column's value, deliberately zeroed in memory before the call, is copied back by `ReloadAll`. Point (c) is what actually exercises the column-copy loop itself; without it, deleting the loop entirely would still pass the test.

`ReloadAll` is used rather than `UpdateAll` because `ReloadAll` has no dialect override, so the test exercises `copyMatchingRows` on every dialect.

There is no assertion on `.C` in the factory output templates: they cannot tell statically whether the counts plugin is enabled, so referring to `obj.C` there would fail to compile for anyone who disables it. `.R` and `.C` are now protected by the exact same mechanism — no assignment statement targets either — so the `.R` assertion covers the mechanism as a whole.

However, bob's own driver test generation templates (`test/gen/templates/factory/{sqlite,psql,mysql}/bobfactory_runtime.bob_test.go.tpl`) are generated with `plugins.PresetAll`, where counts are always enabled, and already reference `.C` directly (e.g. `user.C.Videos`). This PR adds `TestUserSliceReloadAllKeepsCounts` to all three of those files, which fails before this fix (counts silently dropped by `ReloadAll`) and passes after it. These three files also already contained slice-level tests that predate this PR — `TestUserSliceReloadAll` (`ReloadAll`) and `TestVideoSliceDeleteAll` (`DeleteAll`) — none of which break from this change, since they only check column values, not `.R`/`.C`.

## Checklist

- [x] Both branches of `copyMatchingRows` (map fast path and nested-loop fallback) updated.
- [x] Added a regression test template covering the fixed behavior.
- [x] Updated CHANGELOG.md under `[Unreleased]` with a `BREAKING` note.

---

Note: #758 also adds an entry under `[Unreleased]`. If that PR lands first, this entry may need a trivial rebase.


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **BREAKING:** Fixed the generated `<Table>Slice.copyMatchingRows` (used by `UpdateAll`, `ReloadAll`, and the `UpdateMod`/`DeleteMod`/`MergeMod` loaders) dropping relationship and plugin caches when refreshing a slice in place. It used to swap each slice element for a freshly scanned model after copying over only `.R`, silently losing `.C` and any other non-column field. It now copies just the column fields onto the existing model instead, so the slice keeps its original pointers and all its cached fields. See [#759](https://github.com/stephenafamo/bob/pull/759) for the full rationale.
```
stephenafamo: Not sure this is a "Breaking" change

stephenafamo (commented):
Also, doesn't this also affect the singular model's `Update` and `Reload` method?

sandonemaki:
Yes — the singular Update/Reload had exactly the same problem, and that's what #758 fixes (this PR is its slice-side counterpart; the two together cover both paths). #758 applies the same "copy only the column fields" approach to 005_one_methods.go.tpl.

CHANGELOG.md
sandonemaki: Fair point — relying on the old behavior (slice elements being swapped for fresh pointers, or .All() results carrying .R) is unusual enough that this reads better as a plain fix. I've dropped the BREAKING label and retitled the PR fix:, keeping a short note about the pointer-identity change in the entry. #758 currently carries the same label — happy to downgrade that one too if you agree.

sandonemaki:
@stephenafamo 
Yes — the singular Update/Reload had exactly the same problem, and that's what #758 fixes (this PR is its slice-side counterpart; the two together cover both paths). #758 applies the same "copy only the column fields" approach to 005_one_methods.go.tpl.

sandonemaki:
@stephenafamo 
I hope you're doing well. I wanted to check in on the timeline for two PRs I have open:

#758 — fix!: copy only the column fields in the generated single-model Update and Reload
#759 — fix: copy only the column fields in the generated slice copyMatchingRows

Both are ready for review on my side, with the checklist items done. Would it be possible to get them merged sometime in August? We're planning around this on our end, so having a rough idea of the timing would help a lot.

If there's anything you'd like me to change or clarify, I'm happy to address it right away. No pressure if the schedule doesn't allow it — I'd just appreciate knowing either way.

Thanks for your time and for maintaining bob.

Best regards,

stephenafamo:
Kindly fix the merge conflicts.

sandonemaki:
@stephenafamo 
I've resolved the conflict.
Please check it.

stephenafamo:
There seems to be quite a number of additional changes that I'm not sure are related to the PR

sandonemaki:
@stephenafamo 
Apologies for the noise — I mistakenly pushed two repository-wide formatting commits to this branch earlier. They were unrelated to this PR, so I have removed them with a force-push, and the PR now contains only its original changes.

About the lint failure: it is not caused by the changes in this PR. This PR only touches generator templates and CHANGELOG.md, but golangci-lint (gofumpt with extra-rules: true) reports formatting violations in 25 files that this PR does not modify — under gen/bobgen-psql/driver/parser/, gen/bobgen-mysql/driver/parser/, gen/bobgen-sqlite/driver/parser/, dialect/{mysql,psql,sqlite}/, gen/drivers/, etc. These violations already exist in the code this branch inherits from upstream main, so lint fails regardless of this PR's diff. Note that golangci-lint only reports a few issues per run, so the failure surfaces different files each time.

Running gofumpt -extra -l . at the repository root reproduces the full list. Since fixing them is outside the scope of this PR, the formatting fixes are not included here.

sandonemaki:
@stephenafamo 
Apologies for the noise — I mistakenly pushed two repository-wide formatting commits to this branch earlier. They were unrelated to this PR, so I removed them with a force-push, and this PR now contains only its original changes.

About the lint failure: it is not caused by this PR. The lint workflow uses golangci-lint-action with version: latest, and golangci-lint moved from v2.12.2 to v2.13.2 (released 2026-08-27), whose newer gofumpt flags 25 pre-existing files — none of which this PR touches. This is why lint on main itself has also been failing since 519ac21. Since golangci-lint caps identical messages (max-same-issues), each run only surfaces 3 of them, making the failure look different every time.

I have opened #765 with the formatting fixes (formatting only, no logic changes). Could you merge #765 first? After that, re-running the checks here (or updating the branch) should make lint pass on this PR without any further changes.

PR #765, opened 2026-08-30
fix: gofumpt formatting for golangci-lint v2.13

sandonemaki:
The lint workflow on main has been failing since 519ac21 (merge of #758). This is not caused by that PR: lint.yml runs golangci-lint-action with version: latest, and the run on the merge commit picked up golangci-lint v2.13.2 (released 2026-08-27) instead of v2.12.2. The newer bundled gofumpt flags 25 pre-existing files as not properly formatted. Since golangci-lint caps identical messages (max-same-issues), each run only surfaces 3 of them, which makes the failure look different every time.

This PR applies gofumpt -extra -w (matching the gofumpt: extra-rules: true setting in .golangci.yml) to those 25 files. Formatting only — no logic changes.

Note: v2.13 also warns that extra-rules is deprecated in favor of extra.group-params. I have left the config untouched, as that migration seems like a separate decision.

### DIFF
--- a/dialect/mysql/select_test.go
+++ b/dialect/mysql/select_test.go
@@ -93,16 +93,19 @@ func TestSelect(t *testing.T) {
 				GROUP BY status`,
 			Query: mysql.Select(
 				sm.Columns("status", mysql.F("avg", "difference")),
-				sm.From(mysql.Select(
-					sm.Columns(
-						"status",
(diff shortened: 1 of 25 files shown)


sandonemaki:
@stephenafamo 
Opened this to fix the main lint failure caused by the golangci-lint v2.13 auto-upgrade (details in the description). Merging this first would let currently-failing PRs like #759 pass with just a re-run.

